PR #135, opened 2019-05-05
Attempt to implement readOnly field.

megasoft78:
Hi,
I tried to implement the readOnly field and it seems to partially work.
It automatically skip to create the read only column on the create of the table and for insert/update/delete but it's not there for mapper.
Could you please help me to understand what I'm doing wrong?

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,10 +44,10 @@ This package is still in an early phase and the API will likely change.
     dependencies:
       flutter:
         sdk: flutter
-      floor: ^0.4.1
+      floor: ^0.4.2
     
     dev_dependencies:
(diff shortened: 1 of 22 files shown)


vitusortner:
Hi! I'm happy to see your interest and commitment on adding features and improving the library. Unluckily, I'm quite busy right now. I'll prepare some information about the insights of this library (structure, components, etc.) and will come back to you soon!

megasoft78:
@vitusortner Hi,
Don't worry. I was able to make it work. The only problem at the moment is that I have to put the readOnly field in the default constructor to work.
Any idea why?

vitusortner:
Closed by #248

Issue #164, opened 2019-06-30
primary key is assumed to be int for update and delete


North101:
As the title. Everywhere else the primary key is fine but when using commands annotated with @update or @delete they error because it tries to cast the value to int when its not.

North101:
I changed `final int primaryKey` to `final dynamic primaryKey` and `<int>[primaryKey]` to `<dynamic>[primaryKey]` in update_adapter.dart and deletion_adapter.dart and it works fine.

I'm still very new to dart and unfamiliar with the project so I won't open a PR as i'm unsure of any unknown consequences of that change.

MahdiPishguy:
@North101  witch version you are using now?

North101:
I was using `floor: ^0.7.0`. I'm now using my own fork with the following commit: https://github.com/North101/floor/commit/23392ed2f162990cef57d47e0d83388b1ce75d12

vitusortner:
Hi @North101! The changes you've done look fine. Would you mind opening a PR so we could merge these into the library?

North101:
Done. You may also want to restrict PrimaryKey(autoGenerate=true) to ints (if possible)

dkaera:
Hello guys! First I did the same fix, but I found a new one. Update/Delete adapter uses just first item from the primary key list. I have created PR with the fix. Could you review it and leave comments if it's needed. Thanks. 

https://github.com/vitusortner/floor/pull/181

PR #167, opened 2019-07-01
Change delete and update to allow string primary keys

North101:
PR for #164


noordawod:
This is a very good patch! Any plans to merge it @vitusortner?

dkaera:
This fix allows supporting just String primary keys, but I found the new issue. Update and delete adapters doesn't work correctly with group keys. I'll create PR with a fix, which will support `dynamic` primary keys, both single and group.

noordawod:
That would be awesome, @dkaera! Thanks.

vitusortner:
A complete fix for the problem got introduced by #181

PR #181, opened 2019-08-27
Allow int and string (composite) primary keys

### LINKED ISSUE #164: primary key is assumed to be int for update and delete

dkaera:
Closes #164


vitusortner:
@dkaera  Thank you for your contribution! I've just added test for the new functionality and changed the naming a bit.

dkaera:
> @dkaera Thank you for your contribution! I've just added test for the new functionality and changed the naming a bit.

You're welcome. I'm glad to help.

PR #182, opened 2019-08-27
Add optional callback functions when opening database

noordawod:
This is backward-compatible.


noordawod:
This is backward-compatible.

dkaera:
Please, add `floor_generator/lib/generated/i18n.dart (126)` to `.gitignore` . It's an autogenerated file.

noordawod:
> Please, add `floor_generator/lib/generated/i18n.dart (126)` to `.gitignore` . It's an autogenerated file.

Done.

noordawod:
Travis seems to be down or something...

noordawod:
`The command "./tool/travis.sh dartfmt" exited with 1.`

Travis is not feeling well :/ but the code now passes tests if you do it offline: `pub run test`

dkaera:
> `The command "./tool/travis.sh dartfmt" exited with 1.`
> 
> Travis is not feeling well :/ but the code now passes tests if you do it offline: `pub run test`

As I see in CI log, you need to run `flutter format ./floor_generator/` or navigate to `floor_generator` directory and run `flutter format .` from Terminal to apply Dart format for `floor_generator` module.

noordawod:
> > `The command "./tool/travis.sh dartfmt" exited with 1.`
> > Travis is not feeling well :/ but the code now passes tests if you do it offline: `pub run test`
> 
> As I see in CI log, you need to run `flutter format ./floor_generator/` or navigate to `floor_generator` directory and run `flutter format .` from Terminal to apply Dart format for `floor_generator` module.

Ugh, thanks for the tip. Retrying...

noordawod:
Travis stopped?

https://github.com/travis-ci/travis-ci/issues/10204

floor_generator/lib/writer/database_writer.dart
```diff
@@ -94,24 +94,50 @@ class DatabaseWriter implements Writer {
...
+      ..named = true
+      ..type = refer('sqflite.OnDatabaseVersionChangeFn'));
+
     return Method((builder) => builder
       ..name = 'open'
       ..returns = refer('Future<sqflite.Database>')
       ..modifier = MethodModifier.async
       ..requiredParameters.addAll([nameParameter, migrationsParameter])
+      ..optionalParameters
+          .addAll([onConfigureParameter, onCreateParameter, onUpgradeParameter])
```
vitusortner: Please add a comma after `onUpgradeParameter` and reformat the file in order to have a proper line break.

```suggestion
          .addAll([onConfigureParameter, onCreateParameter, onUpgradeParameter,])
```

vitusortner:
Hi @noordawod thanks for your contribution! What's the reasoning for allowing these additional functions? I see some problems with opening the possibility of hooking into the migration (`onUpgrade`) and table creation (`onCreate`) process.

noordawod:
> Hi @noordawod thanks for your contribution! What's the reasoning for allowing these additional functions? I see some problems with opening the possibility of hooking into the migration (`onUpgrade`) and table creation (`onCreate`) process.

Hi @vitusortner.

My use case: I needed to know when the database is first created (so show an onboarding screen), then when an upgrade is going to happen (show a progress indicator), and when the database is up to speed -- and then rush the user to the main screen. I see many people wanting this use case, but maybe it's just me :)

About the callbacks: exposing the database instance could potentially be a problem *if* it isn't exposed in the first place, but I believe it is. If you do think this may cause trouble, then maybe just allow callbacks without the database instance? (new typedefs)

Let me know what you think.

noordawod:
@vitusortner 

Would you like having callbacks without access to the SQLite database? I can do this modification if it fits better with the architecture you want.

vitusortner:
I gave it a second thought and now think, that there is completely valid use cases for exposing the database.

### Implementation
I would prefer passing the callbacks via the generated `DatabaseBuilder` (similar to how migrations are forwarded) instead of using the `DatabaseBuilder.build` function. An example usage is shown in the following snippet.

The callback parameters `onCreate`, `onOpen`, `onUpgrade` should be named arguments.

```dart
final callback = Callback(
   onCreate: (database, version) {},
   onOpen: (database) {},
   onUpgrade: (database, startVersion, endVersion) {},
);

final database = await $FloorTestDatabase
    .addCallback(callback)
    .build();
```

`Callback` could be as simple as the following and should be located at `floor/lib/src/callback.dart`.

```dart
class Callback {
  final FutureOr<void> Function(Database database, int version) onCreate;
  final FutureOr<void> Function(Database database) onOpen;
  final FutureOr<void> Function(
    Database database,
    int startVersion,
    int endVersion,
  ) onUpgrade;

  Callback({this.onCreate, this.onOpen, this.onUpgrade});
}
```

Using `FutureOr` opens up the possibility of either using Dart's asynchronous capabilities in case they are required or to just execute synchronous code.

What made you decide on executing the passed functions before the migrations and table creations happen?

noordawod:
@vitusortner 

This should work nicely too! What's your eta for this?

The decision stems from old habbits (from the JavaScript/jQuery) world: the handler may return a `false` to cancel an event (click for example). But in our use case, it could be beneficial that these callbacks happen after Floor have done its work. Up to you.

vitusortner:
If you'd find time for the implementation, I'd welcome your contribution! Based on the changes you've done, you know which places of the code base to touch.

Let's try with putting the callbacks after the initialization has happened.

noordawod:
On it. I'll push to this pr.

noordawod:
@vitusortner 
Working nicely in my project!

floor/lib/src/callback.dart
```diff
@@ -0,0 +1,25 @@
+import 'dart:async';
+
+import 'package:sqflite/sqflite.dart' as sqflite;
```
vitusortner: Library prefix `as sqflite` can be removed as there are no naming collisions in this class.

floor_generator/lib/writer/database_builder_writer.dart
```diff
@@ -24,6 +24,11 @@ class DatabaseBuilderWriter extends Writer {
       ..modifier = FieldModifier.final$
       ..assignment = const Code('[]'));
 
+    final callbackField = Field((builder) => builder
+      ..name = '_callback'
+      ..type = refer('Callback')
+      ..modifier = FieldModifier.var$);
```
vitusortner: The modifier doesn't have to be defined in this case.

floor_generator/lib/writer/database_builder_writer.dart
```diff
@@ -41,47 +46,46 @@ class DatabaseBuilderWriter extends Writer {
...
         final database = _\$$_databaseName();
         database.database = await database.open(
           name ?? ':memory:',
           _migrations,
-          onConfigure: onConfigure,
-          onCreate: onCreate,
-          onUpgrade: onUpgrade,
+          _callback,
         );
+
```
vitusortner: Lets keep the generated code as small as possible.

- [ ]  Remove empty line

floor_generator/lib/writer/database_writer.dart
```diff
@@ -94,55 +94,44 @@ class DatabaseWriter implements Writer {
...
       ..name = 'open'
       ..returns = refer('Future<sqflite.Database>')
       ..modifier = MethodModifier.async
       ..requiredParameters.addAll([nameParameter, migrationsParameter])
-      ..optionalParameters.addAll([
-        onConfigureParameter,
-        onCreateParameter,
-        onUpgradeParameter,
-      ])
+      ..optionalParameters.addAll([callbackParameter])
```
vitusortner: A list is not required here. You could use `add(callbackParameter)`.

floor_generator/lib/writer/database_writer.dart
```diff
@@ -94,55 +94,44 @@ class DatabaseWriter implements Writer {
...
             onConfigure: (database) async {
-              if (onConfigure != null) {
-                await onConfigure(database);
-              }
               await database.execute('PRAGMA foreign_keys = ON');
             },
+            onOpen: (database) async {
+              if (callback?.onOpen != null) {
+                await callback.onOpen(database);
+              }
```
vitusortner: This could be simplified with the following. Same applies for the other callbacks.

```dart
await callback?.onOpen?.call(database);
```

Do you have an idea how to only use `await` in case `onOpen` returns a `Future`?

.gitignore
```diff
@@ -31,6 +31,8 @@
 .pub/
 build/
 pubspec.lock
+**/generated/i18n.dart
+**/values/*.arb
```
vitusortner: Are these excludes still required?

noordawod: Probably not, will remove them.

floor_generator/lib/writer/database_writer.dart
noordawod: Maybe something like this would work, untested:

```dart
final result = callback?.onOpen?.call(database);
if (result is Future) {
  await result;
}
```

noordawod: But I believe having an `await` has no side effects if the callback function is not `async`.

vitusortner:
Looking good so far. I'll add some documentation to the project's README files on the new callbacks and merge the PR after that. Thanks for your assistance!

noordawod:
Sounds good! Maybe also add a test for the new functionality?

vitusortner:
Yes, it would be beneficial to verify that each callback is invoked when it should be. Unluckily, I currently see no simple way of doing that.

noordawod:
Coming to `pub.dev` anytime soon? :)

vitusortner:
Is now released.

PR #184, opened 2019-09-01
Update README with clear package import instructions

afzaal-ahmad-zeeshan:
It is important to mention that these imports are required; I used Floor but I chose to develop my own classes and had to waste a huge amount of time finding the solution.

This comment can catch eye of the reader and they will know the purpose of these specific imports.


vitusortner:
Thanks for your contribution! I'll also update the copy of the README you've edited in order to have them in sync.

PR #189, opened 2019-09-11
Let users choose to add @override to open() method.

noordawod:
This is handy once the generated classes reside in a different package, then `_$AppDatabase` becomes unreachable. So the abstract `AppDatabase` class would have the signature of `open()` while the generated code adds the `@override` annotation so to keep the linter happy.

### DIFF
--- a/floor_annotation/lib/src/database.dart
+++ b/floor_annotation/lib/src/database.dart
@@ -8,6 +8,13 @@ class Database {
   /// The entities the database manages.
   final List<Type> entities;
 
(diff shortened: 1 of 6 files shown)


noordawod:
Hmm, or better add the signature in `FloorDatabase` class and have `open()` always annotated with `@override`?

noordawod:
@vitusortner 

Superseded by #190 -- decide which pill is better :)

noordawod:
Refactored my use case, no need for this anymore.

PR #190, opened 2019-09-11
Add open() method signature to FloorDatabase class.

noordawod:
This is handy once the generated classes reside in a different package, then `_$AppDatabase` becomes unreachable. So the abstract `AppDatabase` class would have the signature of `open()` while the generated code adds the `@override` annotation so to keep the linter happy.

### DIFF
--- a/floor/lib/src/database.dart
+++ b/floor/lib/src/database.dart
@@ -3,6 +3,8 @@ import 'dart:async';
 import 'package:meta/meta.dart';
 import 'package:sqflite/sqflite.dart' as sqflite;
 
+import 'migration.dart';
+
 /// Extend this class to enable database functionality.
 abstract class FloorDatabase {
   /// [StreamController] that is responsible for notifying listeners about changes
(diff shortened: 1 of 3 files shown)


vitusortner:
Hi @noordawod, what's your use case for this? It should be the same as opening the database by using the database builder.

noordawod:
@vitusortner 

Tried explaining my usecase in the previous comment, so I'll try again.

I architectured my app in such a way that the data model lives in its own module. The data model contains Floor and all its generated code. In this way, however, `_$AppDatabase` is basically inaccessible. I could add a signature for `open()` in the abstract `AppDatabase` class annotated with `@Database`, but there should be an `@override` annotation there too.

The reason this is good is that I could just reference `FloorDatabase` instead of `AppDatabase` in method/function signatures, like so:

```dart
Future<void> ensureDatabaseOpen(covariant FloorDatabase db) {
  return db.open(...);
}
```

There's no need to specifically request that the passed parameter is a specific database class (like AppDatabase). It makes more sense when creating multiple schemas, for example.

Lastly, I could argue that a class that has a `close()` method should also have an `open()` counterpart.

noordawod:
What do you say @vitusortner?

noordawod:
What do you say @vitusortner?

noordawod:
Refactored my use case, no need for this anymore.

PR #192, opened 2019-09-14
Make IN clauses work with strings

### LINKED ISSUE #171: [BUG] Incorrect query compiling
Hello.

Bug in query:
@Query('SELECT * FROM Dog WHERE name IN (:names)')
Future<List<Dog>> getAllByNames(List<String> names);

When arguments are substituted in query they are not wrapped in apostrophe ( ' ).

Compiled query for call getAllByNames(["name1", "name2]) will be:

'SELECT * FROM Dog WHERE name IN (name1, name2)'

bur right query is:

'SELECT * FROM Dog WHERE name IN ('name1', 'name2')'

vitusortner:
IN clauses used to only work with integers, as string values were not surrounded by quotes.

Previous: `'SELECT * FROM Dog WHERE name IN (name1, name2)'`
Fixed: `'SELECT * FROM Dog WHERE name IN ('name1', 'name2')'`

Resolves #171

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,7 @@ pubspec.lock
 **/ios/Flutter/flutter_assets/
 **/ios/ServiceDefinitions.json
 **/ios/Runner/GeneratedPluginRegistrant.*
+**/flutter_export_environment.sh
 
 # Exceptions to above rules.
 !**/ios/**/default.mode1v3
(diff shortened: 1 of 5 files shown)


noordawod:
Ahh, this is great! In my existing DAOs I had to write private implementations with @Query and public ones that escapes the entries. Good change, thanks!

PR #203, opened 2019-10-06
Support bool values in a Map stored as a real bool (not int).

noordawod:
This compliments a pr #178 when the Map is retrieved from source other than SQLite (API response, JSON file, etc.)

### DIFF
--- a/floor_generator/lib/processor/entity_processor.dart
+++ b/floor_generator/lib/processor/entity_processor.dart
@@ -257,7 +257,7 @@ class EntityProcessor extends Processor<Entity> {
     final String parameterValue,
   ) {
     if (isBool(parameterType)) {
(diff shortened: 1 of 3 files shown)


noordawod:
No idea what's failing, Codedev is so obscure at hiding details.

noordawod:
@vitusortner 

I got another pr waiting that unifies repetitive code and adds an exporter function (like the query mapper, but in the opposite direction: to map an entity to a Map). Could you find few moments to go over this pr?

noordawod:
Complimentary Entity-to-Map exporter has landed: #204. Code adjusted to use it (Querying, Inserting, Deleting).

noordawod:
Anyone wanting to have this feature can use my fork which I'll maintain to always sync with upstream:develop: https://github.com/noordawod/floor/tree/develop-noordawod

noordawod:
This one too was supposed to cut on duplication of field names, I don't understand why it wasn't merged.

PR #204, opened 2019-10-08
Add a reusable exporter function (Entity to Map).

noordawod:
This relies and compliments pr #203.


noordawod:
Anyone wanting to have this feature can use my fork which I'll maintain to always sync with upstream:develop: https://github.com/noordawod/floor/tree/develop-noordawod

noordawod:
Too bad this wasn't merged in, it'd have made my life much easier rather than having duplicates of the same fields that already exist in the db schema.

PR #233, opened 2020-01-17
Exclude static fields from entity mapping

### LINKED ISSUE #231: Static fields should not be included in the mapping
Currently the static fields are included in
1) Generated CREATE TABLE sql, and
2) Row data mapper.

As example,
```
class Entity {
   static const tableName = 'entity';

   @PrimaryKey(autoGenerate: true)
   final int id;
   final double price;
   Entity(this.id, this.price);
}
```
The generated CREATE TABLE sql for this class looks like
`CREATE TABLE IF NOT EXISTS Entity  (tableName  TEXT, id INTEGER PRIMARY KEY AUTOINCREMENT, price REAL)`

And the row mapper is totally wrong:
```
static final _entityMapper = (Map<String, dynamic> row) => Entity(
      row['tableName'] as int,
      row['id'] as double);
```

finger-software:
The static fields are excluded from the field list for mapping.

Closes #231


vitusortner:
Thanks for your contribution! I'll add tests for the new code to then merge it.

PR #259, opened 2020-02-13
Feature: Add support for ByteArrays/Blobs

mqus:
This connects the dart type Uint8List to the SQLite data type BLOB and adjusts some tests for that.
Fixes #229.

I'm a newbie to Dart/Flutter so please make sure that I did nothing wrong. I tested it locally but the test didn't get through for an unrelated reason (I assume some race condition?) But I can be wrong.

```
$ flutter test
00:02 +68 -1: [...]/floor/floor/test/integration/database_test.dart: database tests stream queries insert change find person by id as stream [E]                                                       
  Bad state: Cannot add new events after calling close
  dart:async                                          _BroadcastStreamController.add
  package:floor/src/adapter/query_adapter.dart 68:38  QueryAdapter.queryStream.executeQueryAndNotifyController
  ===== asynchronous gap ===========================
  dart:async                                          _asyncThenWrapperHelper
  package:floor/src/adapter/query_adapter.dart        QueryAdapter.queryStream.executeQueryAndNotifyController
  package:floor/src/adapter/query_adapter.dart 76:55  QueryAdapter.queryStream.<fn>
  ===== asynchronous gap ===========================
  dart:async                                          _ForwardingStream.listen
  package:floor/src/adapter/query_adapter.dart 75:10  QueryAdapter.queryStream
  integration/database.g.dart 167:26                  _$PersonDao.findPersonByIdAsStream
  integration/database_test.dart 274:36               main.<fn>.<fn>.<fn>.<fn>
  
00:03 +77 -1: Some tests failed.  
```

Issue #130, opened 2019-05-04
Allow query to return objects not entities


megasoft78:
Hi,
I'm using your amazing package for my first app on Flutter and I'm having an issue.
I want to query the data and return the result of these queries into some objects that are not table.

For instance if I want to have the result of a group by query like this one:

```
Select R.roomName, count(*) as numberOfStudents
from tblRooms R
Inner Join tblStudents S on (S.roomId=R.roomId)
Group By R.roomName
```

I need the possibility to setup a class like this:

```
@View()
class RoomStudentsCountView {
  String roomName;
  int numberOfStudents;

  RoomStudentsCountView (this.roomName, this.numberOfStudents);
}
```

I replaced the @Entity with @View as alternative to describe this type of entity.

I also find sometime that I need the possibility to define field that are not column of the table but just calculated field.
Here's an example:

```
@Entity(tableName: "tblRooms")
class RoomEntity {
  @PrimaryKey(autoGenerate: true)
  int roomId;
  String roomName;

  @CalculatedField()
  int numberOfStudents;

  RoomEntity (this.roomId, this.roomName, this.numberOfStudents);
}
```

I can try to help by looking at the code and try to adapt it but I'm not sure I have enough experience to make it right.

I hope you can help on this. :)

vitusortner:
Hi @megasoft78! I'm happy to see your interest in the package. To your first request: It's planned to support returning objects for queries that are not entities. #33 holds some more information about this feature request. To your second request: This behavior is also planned. #12 describes the feature.

I've seen you forked the repository. Do you plan on implementing the behavior? Would be great to see a PR!

megasoft78:
Hi @vitusortner,
I'm new to Flutter and Dart as I mainly work on C# but I'm looking to implement #12.
My idea is to add a new annotation to ColumnInfo : readonly. I did that but I'm learning where to change the code in order to achieve this feature.
I'm looking for all the CRUD operations and skip the field with readonly annotation for Create, Update and Delete.
Could you give me some info on where to look and how is structured?

mqus:
Hi! the current state of android development pushed me to flutter and yesterday I found your library which seems to perfectly replace room. I would really like to see the `@DatabaseView`s in floor like I had them in room:
```java
@DatabaseView("SELECT collection, artist AS name, group_concat(DISTINCT genre) AS genres, count(*) AS count FROM files GROUP BY artist, collection")
public class Artist {
	String collection;
	String name;
	String genres;
	int count;
}
[...]

PR #262, opened 2020-02-15
Feature: Add support for @DatabaseView annotations

mqus:
This adds support for @DatabaseView annotations known from room. One test was added, further testing is needed. Implements #130. I tried to do minimal adjustments to the codebase while copying the handling of entities. I took over the Field handling from there and then just ignored the nullable property by implementing the createStatement function differently. 

As I already said in #259, I may have no idea what I'm doing, so please tell me if I did something wrong or if I can do something better.

I'll try to add some more tests before this is ready to merge. Are there some specific things you would like to see tested?

### DIFF
--- a/floor/test/integration/database_test.dart
+++ b/floor/test/integration/database_test.dart
@@ -304,36 +304,37 @@ void main() {
         test('query view with exact value', () async {
           final person = Person(1, 'Frank');
           await personDao.insertPerson(person);
-          final expectedName = Name('Frank');
+
           final actual = await nameDao.findExactName('Frank');
 
-          expect(actual, equals(expectedName));
+        final expected = Name('Frank');
+          expect(actual, equals(expected));
         });
 
         test('query view with LIKE', () async {
-          final person = Person(1, 'Leo');
+          final persons = [Person(1, 'Leo'), Person(2, 'Frank')];
+          await personDao.insertPersons(persons);
+
           final dog = Dog(1, 'Romeo', 'Rome', 1);
-          await personDao.insertPerson(person);
-          await personDao.insertPerson(Person(2, 'Frank'));
           await dogDao.insertDog(dog);
 
-          final expectedNames = [Name('Leo'), Name('Romeo')];
           final actual = await nameDao.findNamesLike('%eo');
 
-          expect(actual, equals(expectedNames));
+          final expected = [Name('Leo'), Name('Romeo')];
+          expect(actual, equals(expected));
         });
 
         test('query view with all values', () async {
-          final person = Person(1, 'Leo');
+          final persons = [Person(1, 'Leo'), Person(2, 'Frank')];
+          await personDao.insertPersons(persons);
+
           final dog = Dog(1, 'Romeo', 'Rome', 1);
-          await personDao.insertPerson(person);
-          await personDao.insertPerson(Person(2, 'Frank'));
           await dogDao.insertDog(dog);
 
-          final expectedNames = [Name('Frank'), Name('Leo'), Name('Romeo')];
           final actual = await nameDao.findAllNames();
 
-          expect(actual, equals(expectedNames));
+          final expected = [Name('Frank'), Name('Leo'), Name('Romeo')];
+          expect(actual, equals(expected));
         });
       });
     });
--- a/floor/test/integration/database_test.dart
+++ b/floor/test/integration/database_test.dart
@@ -6,11 +6,9 @@ import 'package:sqflite_ffi_test/sqflite_ffi_test.dart';
 
 import '../test_util/extensions.dart';
 import 'dao/dog_dao.dart';
-import 'dao/name_dao.dart';
 import 'dao/person_dao.dart';
 import 'database.dart';
 import 'model/dog.dart';
-import 'model/name.dart';
 import 'model/person.dart';
 
 void main() {
@@ -21,7 +19,6 @@ void main() {
     TestDatabase database;
     PersonDao personDao;
     DogDao dogDao;
-    NameDao nameDao;
 
     setUp(() async {
       final migration1to2 = Migration(1, 2, (database) {
@@ -36,7 +33,6 @@ void main() {
 
       personDao = database.personDao;
       dogDao = database.dogDao;
-      nameDao = database.nameDao;
     });
 
     tearDown(() async {
@@ -297,44 +293,6 @@ void main() {
           final expectedPersons =
               persons.where((person) => person.name.contains('a'));
           expect(actual, equals(expectedPersons));
-        });
-      });
-
-      group('Query View', () {
-        test('query view with exact value', () async {
-          final person = Person(1, 'Frank');
-          await personDao.insertPerson(person);
-
-          final actual = await nameDao.findExactName('Frank');
-
-        final expected = Name('Frank');
-          expect(actual, equals(expected));
-        });
-
-        test('query view with LIKE', () async {
-          final persons = [Person(1, 'Leo'), Person(2, 'Frank')];
-          await personDao.insertPersons(persons);
-
-          final dog = Dog(1, 'Romeo', 'Rome', 1);
-          await dogDao.insertDog(dog);
-
-          final actual = await nameDao.findNamesLike('%eo');
-
-          final expected = [Name('Leo'), Name('Romeo')];
-          expect(actual, equals(expected));
-        });
-
-        test('query view with all values', () async {
-          final persons = [Person(1, 'Leo'), Person(2, 'Frank')];
-          await personDao.insertPersons(persons);
-
-          final dog = Dog(1, 'Romeo', 'Rome', 1);
-          await dogDao.insertDog(dog);
-
-          final actual = await nameDao.findAllNames();
-
-          final expected = [Name('Frank'), Name('Leo'), Name('Romeo')];
-          expect(actual, equals(expected));
         });
       });
     });
--- a/floor/test/integration/database_test.dart
+++ b/floor/test/integration/database_test.dart
@@ -6,9 +6,11 @@ import 'package:sqflite_ffi_test/sqflite_ffi_test.dart';
 
 import '../test_util/extensions.dart';
 import 'dao/dog_dao.dart';
+import 'dao/name_dao.dart';
 import 'dao/person_dao.dart';
 import 'database.dart';
 import 'model/dog.dart';
+import 'model/name.dart';
 import 'model/person.dart';
 
 void main() {
@@ -19,6 +21,7 @@ void main() {
     TestDatabase database;
     PersonDao personDao;
     DogDao dogDao;
+    NameDao nameDao;
 
     setUp(() async {
       final migration1to2 = Migration(1, 2, (database) {
@@ -33,6 +36,7 @@ void main() {
 
       personDao = database.personDao;
       dogDao = database.dogDao;
+      nameDao = database.nameDao;
     });
 
     tearDown(() async {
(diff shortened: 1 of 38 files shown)


mqus:
The test that's currently failing is the same as in #259 and as far as I can tell, this fails even on the current developer branch. I assume that some underlying framework (dart?) got changed but I couldn't find the issue with my very limited knowledge of `Stream`s.

vitusortner:
At a first glance, this looks great 🎉 Thanks for your contribution! I'll review the full PR as soon as possible.

mqus:
Some remarks about hacks I did and shouldn't change back myself right now:
- floor_generator relies on floor_annotation via `path: ../floor_annotation` because it needs the new annotations. This has to be changed before the next release but can't be changed right now because nothing would work anymore.
- the same for floor, which relies on floor_generator and floor_annotation.

I also messed up my branches a bit when starting out so view_processor.dart already checks for Uint8List, even if the whole rest of the code can't do anything with it. If you want to merge this PR before #259, I'll remove it. If #259 comes first, I'll rebase and use the proper getter implemented there.

vitusortner:
> * floor_generator relies on floor_annotation via `path: ../floor_annotation` because it needs the new annotations. This has to be changed before the next release but can't be changed right now because nothing would work anymore.
> * the same for floor, which relies on floor_generator and floor_annotation.

No problem. Normally, I only change the paths of the library's packages to point to the versions available on pub while creating a release. I must have forgotten to revert it to referring the local version. You can leave it as it is.

> I also messed up my branches a bit when starting out so view_processor.dart already checks for Uint8List, even if the whole rest of the code can't do anything with it. If you want to merge this PR before #259, I'll remove it. If #259 comes first, I'll rebase and use the proper getter there.

Let's merge #259 before.

mqus:
I rebased the PR onto the current develop branch and altered `view_processor.dart` as written before. 
I know that an integration test for the Views is still missing and maybe more, do you have any suggestions on what/how should be tested?

mqus:
So, I rebased to the current develop and added some more tests (CodeCov seems to work even for PRs now, nice :) )

There is only one open question for me now: There seems to be a mechanism which checks if a table has been altered and then notifies the watchers of the concerning tables. Is that correct? If yes, do you expect this to also work with views? If yes again, then there are three ways to get the tables/entities we have to watch for a specific `DatabaseView`: 

1. We try to detect them in the select query for the view (Seems pretty impossible with all the possibilities to include table names and naming things similarly)
2. Users have to provide the Entities on which a view depends in the `@DatabaseView` annotation. (easy but could lead to an unsafe/leaky way of using the library, since we can't check or guarantee the relation)
3. SQLite/sqflite somehow provides this (I'm not aware of that)
4. Any other Ideas?

Btw, if that watcher mechanism exists, how does it work and how do I access it as a library user?

PS: Ah, right, I should document the DatabaseView usage in the README, or do you want to do it yourself?

mqus:
Thank you! I wont disrupt you then ;). Regarding Streams: Afaict Room seems to take the first option and includes a whole sqlite parser...

floor_generator/lib/processor/query_method_processor.dart
```diff
@@ -42,11 +48,16 @@ class QueryMethodProcessor extends Processor<QueryMethod> {
...
+    final Queryable queryable = _entities.firstWhere(
+            (entity) =>
+                entity.classElement.displayName ==
+                flattenedReturnType.getDisplayString(),
+            orElse: () => null) ??
+        _views.firstWhere(
+            (view) =>
+                view.classElement.displayName ==
+                flattenedReturnType.getDisplayString(),
+            orElse: () => null);
```
vitusortner:
```suggestion
            orElse: () => null); // doesn't return entity nor view
```

floor_generator/lib/processor/query_method_processor.dart
```diff
@@ -42,11 +48,16 @@ class QueryMethodProcessor extends Processor<QueryMethod> {
       returnsStream,
     );
 
-    final entity = _entities.firstWhere(
-        (entity) =>
-            entity.classElement.displayName ==
-            flattenedReturnType.getDisplayString(),
-        orElse: () => null); // doesn't return an entity
+    final Queryable queryable = _entities.firstWhere(
```
vitusortner:
```suggestion
    final queryable = _entities.firstWhere(
```

floor_generator/lib/value_object/dao.dart
```diff
@@ -14,7 +14,7 @@ class Dao {
   final List<UpdateMethod> updateMethods;
   final List<DeletionMethod> deletionMethods;
   final List<TransactionMethod> transactionMethods;
-  final List<Entity> streamEntities;
+  final List<Queryable> streamEntities;
```
vitusortner: We only need entities here for now.
```suggestion
  final List<Entity> streamEntities;
```

floor_generator/test/processor/view_processor_test.dart
```diff
@@ -0,0 +1,274 @@
...
+      
+        Person(this.id, this.name, [this.foo = 'foo']);
+      }
+    ''');
+
+      final actual = ViewProcessor(classElement).process().constructor;
+
+      const expected = "Person(row['id'] as int, row['name'] as String)";
+      expect(actual, equals(expected));
+    });
```
vitusortner: These are basically duplicates from `entity_processor_test`. After migrating duplicate code to `QueryableProcessor`, we could only test this functionality there which would allow us to get rid of this duplication.

floor_generator/lib/processor/view_processor.dart
```diff
@@ -0,0 +1,126 @@
...
+import 'package:floor_generator/misc/annotations.dart';
+import 'package:floor_generator/misc/constants.dart';
+import 'package:floor_generator/misc/type_utils.dart';
+import 'package:floor_generator/processor/error/view_processor_error.dart';
+import 'package:floor_generator/processor/field_processor.dart';
+import 'package:floor_generator/processor/processor.dart';
+import 'package:floor_generator/value_object/field.dart';
+import 'package:floor_generator/value_object/view.dart';
+
+class ViewProcessor extends Processor<View> {
```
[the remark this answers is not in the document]
mqus: I didn't put `_getName()` in the `QueryableProcessor` because it is pretty dependent on the Type (`View` vs `Entity`). But for the rest that's a pretty good idea.

floor_generator/pubspec.yaml
```diff
@@ -18,7 +18,7 @@ dependencies:
   build_config: ^0.4.1+1
   collection: ^1.14.11
   floor_annotation:
-    path: ../floor_annotation/
+    path: ../floor_annotation
```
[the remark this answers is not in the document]
mqus: Does that have a functional difference? Or is this for consistency?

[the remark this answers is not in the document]
vitusortner: Just for consistency 👍

floor/test/integration/database_test.dart
```diff
@@ -295,6 +299,43 @@ void main() {
           expect(actual, equals(expectedPersons));
         });
       });
+
+      group('Query View', () {
```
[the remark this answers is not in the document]
mqus: So I'll move the NameDao, the Name class and copy all other DAOs and entities to a file in `view_test/` and put the tests there, too (like in blob_test)? This seems pretty redundant to me...

I have the following counter proposal: I'll move just the tests (and maybe the database class) into a separate file (or directory) but will still reference the common database objects.
What do you think?

floor_generator/test/processor/view_processor_test.dart
mqus: I don't see a simple way to achieve this, tbh, since `QueryableProcessor` is an abstract class. We would have to instantiate the concrete types of `ViewProcessor` and `EntityProcessor` anyway... But if you have a concrete vision on how to do it, please share and I will try.

floor/test/integration/database_test.dart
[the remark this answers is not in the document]
vitusortner: Sounds great!

floor_generator/test/processor/view_processor_test.dart
vitusortner: We could just create an implementation class of `QueryableProcessor` in the test and asserts its methods for correctness.

```dart
class UnderTest extends QueryableProcessor<Object> {
  UnderTest(ClassElement classElement) : super(classElement);

  @override
  Object process() {
    return null;
  }
}
```

mqus:
I rebased, integrated most of the suggestions and added an exception if a library user will try to use a Dao method which will stream the results of a query on a database view.

floor/test/integration/view_test/view_test.dart
```diff
@@ -0,0 +1,95 @@
...
+import '../dao/dog_dao.dart';
+import '../dao/name_dao.dart';
+import '../dao/person_dao.dart';
+import '../model/dog.dart';
+import '../model/name.dart';
+import '../model/person.dart';
+
+part 'view_test.g.dart';
+
+@Database(version: 2, entities: [Person, Dog], views: [Name])
```
vitusortner: 💡We can get of all the migration related code in this file as we're not testing migrations.

```suggestion
@Database(version: 1, entities: [Person, Dog], views: [Name])
```

floor/test/integration/view_test/view_test.dart
```diff
@@ -0,0 +1,95 @@
...
+    setUp(() async {
+      final migration1to2 = Migration(1, 2, (database) {
+        database.execute('ALTER TABLE dog ADD COLUMN nick_name TEXT');
+      });
+      final allMigrations = [migration1to2];
+
+      database = await $FloorViewTestDatabase
+          .inMemoryDatabaseBuilder()
+          .addMigrations(allMigrations)
+          .build();
```
vitusortner:
```suggestion
      database = await $FloorViewTestDatabase.inMemoryDatabaseBuilder().build();
```

floor_generator/lib/processor/queryable_processor.dart
```diff
@@ -0,0 +1,87 @@
...
+import 'package:analyzer/dart/element/type.dart';
+import 'package:floor_annotation/floor_annotation.dart' as annotations;
+import 'package:floor_generator/misc/annotations.dart';
+import 'package:floor_generator/misc/type_utils.dart';
+import 'package:floor_generator/processor/field_processor.dart';
+import 'package:floor_generator/processor/processor.dart';
+import 'package:floor_generator/value_object/field.dart';
+import 'package:meta/meta.dart';
+
+abstract class QueryableProcessor<T> extends Processor<T> {
```
vitusortner: 💡We could give the generic type some boundaries.

```suggestion
abstract class QueryableProcessor<T extends Queryable> extends Processor<T> {
```

floor_generator/lib/value_object/database.dart
```diff
@@ -26,6 +29,7 @@ class Database {
           classElement == other.classElement &&
           name == other.name &&
           entities == other.entities &&
+          views == other.views &&
           daoGetters == other.daoGetters &&
```
vitusortner: 💡We should start using `ListEquality` for comparing lists in the database class as well.

```suggestion
          const ListEquality<Entity>().equals(entities, other.entities) &&
          const ListEquality<View>().equals(views, other.views) &&
          const ListEquality<DaoGetter>()
              .equals(daoGetters, other.daoGetters) &&
```

floor_generator/lib/value_object/dao.dart
```diff
@@ -1,11 +1,12 @@
 import 'package:analyzer/dart/element/element.dart';
 import 'package:floor_generator/value_object/deletion_method.dart';
-import 'package:floor_generator/value_object/entity.dart';
 import 'package:floor_generator/value_object/insertion_method.dart';
 import 'package:floor_generator/value_object/query_method.dart';
 import 'package:floor_generator/value_object/transaction_method.dart';
 import 'package:floor_generator/value_object/update_method.dart';
 
+import 'entity.dart';
```
vitusortner: 💡We could get rid of all the unnecessary import qualifiers.

```dart
import 'deletion_method.dart';
import 'entity.dart';
import 'insertion_method.dart';
import 'query_method.dart';
import 'transaction_method.dart';
import 'update_method.dart';
```

mqus: Then we should do it everywhere... but I think this should be part of another PR... I'll just put the entity import back where it was for now to make int consistent with the other statements...

floor_generator/test/processor/view_processor_test.dart
mqus: I'll try!

mqus:
I adopted the suggestions from the review and added some documentation to the README files.

tulioccalazans:
@vitusortner  When this  feature will be available as a release?

vitusortner:
We've just created a new release with the version 0.12.0, which makes database views available and includes other features, fixes and improvements.

PR #279, opened 2020-03-11
Ignore Getters&Setters

### LINKED ISSUE #260: Exclude getters and setters from mapping

### LINKED ISSUE #272: How to use ignore annotation?

mqus:
Fixes #272 and fixes #260 by ignoring synthetic fields(getters and setters) entirely.

Should i also remove the `isHashCode` check? it seems redundant now...

Do you want any documentation for this or would you prefer to do it yourself?

### DIFF
--- a/floor/README.md
+++ b/floor/README.md
@@ -395,7 +395,7 @@ class Person {
 ```
 
 ### Ignoring Fields
-The `hashCode` property and all static fields of entities are ignored by default and thus excluded from the library's mapping.
+Getters, setters, operators and all static fields of entities are ignored by default and thus excluded from the library's mapping.
 In case further fields should be ignored, the `@ignore` annotation should be used and applied as shown in the following snippet.
 
 ```dart
@@ -407,6 +407,9 @@ class Person {
 
   @ignore
   String nickname;
+
+  //ignored by default
+  String get combinedName => "$name ($nickname)";
 
   Person(this.id, this.name);
 }
--- a/floor/README.md
+++ b/floor/README.md
@@ -395,7 +395,7 @@ class Person {
 ```
 
 ### Ignoring Fields
-Getters, setters, operators and all static fields of entities are ignored by default and thus excluded from the library's mapping.
+Getters, setters and all static fields of entities are ignored by default and thus excluded from the library's mapping.
 In case further fields should be ignored, the `@ignore` annotation should be used and applied as shown in the following snippet.
 
 ```dart
@@ -408,7 +408,7 @@ class Person {
   @ignore
   String nickname;
 
-  //ignored by default
+  // ignored by default
   String get combinedName => "$name ($nickname)";
 
   Person(this.id, this.name);
--- a/floor_generator/lib/processor/entity_processor.dart
+++ b/floor_generator/lib/processor/entity_processor.dart
@@ -278,6 +278,6 @@ extension on FieldElement {
   bool shouldBeIncluded() {
     final isIgnored = hasAnnotation(annotations.ignore.runtimeType);
     final isHashCode = displayName == 'hashCode';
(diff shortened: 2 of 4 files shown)


floor_generator/lib/processor/entity_processor.dart
```diff
@@ -278,6 +278,6 @@ extension on FieldElement {
   bool shouldBeIncluded() {
     final isIgnored = hasAnnotation(annotations.ignore.runtimeType);
     final isHashCode = displayName == 'hashCode';
```
vitusortner: 💡Yes, let's get rid of this field as you already mentioned.

floor_generator/test/processor/entity_processor_test.dart
```diff
@@ -120,6 +120,50 @@ void main() {
...
+        Person(this.id, this.name);
+        
+        static String foo = 'foo';
+      }
+    ''');
+
+    final actual = EntityProcessor(classElement).process();
+
+    expect(actual.fields.length, equals(2));
+  });
```
vitusortner: 💡 We could get rid of the static field and make the test a bit more explicit as shown in the following. Same applies to the "ignore getter" test.

```suggestion
    test('Ignore setter', () async {
      final classElement = await _createClassElement('''
      @entity
      class Person {
        @primaryKey
        final int id;
      
        final String name;
      
        set printwith(String prefix) => print(prefix+name);

        Person(this.id, this.name);
      }
    ''');

      final actual = EntityProcessor(classElement)
          .process()
          .fields
          .map((field) => field.name)
          .toList();

      expect(actual, equals(['id', 'name']));
    });
```

floor_generator/test/processor/entity_processor_test.dart
```diff
@@ -120,6 +120,56 @@ void main() {
...
+  test('Ignore getter', () async {
+    final classElement = await _createClassElement('''
+      @entity
+      class Person {
+        @primaryKey
+        final int id;
+      
+        final String name;
+      
+        String get label => "" + id + ": " + name
```
vitusortner: 💡Dart's string interpolation is pretty powerful. You could give it a try.

```suggestion
        String get label => '$id: $name';
```

floor/README.md
```diff
@@ -395,7 +395,7 @@ class Person {
 ```
 
 ### Ignoring Fields
-The `hashCode` property and all static fields of entities are ignored by default and thus excluded from the library's mapping.
+Getters, setters, operators and all static fields of entities are ignored by default and thus excluded from the library's mapping.
```
vitusortner: ❓What do you mean by "operators" here?

floor/README.md
```diff
@@ -408,6 +408,9 @@ class Person {
   @ignore
   String nickname;
 
+  //ignored by default
```
vitusortner:
```suggestion
  // ignored by default
```

floor/README.md
mqus: the overridable operators like ==, +, etc.. but this is probably as obvious as other functions... I should remove that.

vitusortner:
I've fixed the failing floor CI job. These changes are now available on the `develop` branch.

floor_generator/test/processor/entity_processor_test.dart
mqus: In this case it is *too* good, because it tries to get the values from outside the class definition:
```
  Failed to load "test/processor/entity_processor_test.dart":
  Unable to spawn isolate: test/processor/entity_processor_test.dart:132:31: Error: Getter not found: 'id'.
          String get label => '$id: $name'
                                ^^
  test/processor/entity_processor_test.dart:132:36: Error: Getter not found: 'name'.
          String get label => '$id: $name'
                                     ^^^^

```
I escaped the dollar signs now.

PR #281, opened 2020-03-13
Support inherited DAO methods

### LINKED ISSUE #258: Dao class is not generated properly with its base class
Hi, I created a base class to have some basic functions for all my Dao classes like that:

```
abstract class BaseDao<T> {  

  @Insert(onConflict: OnConflictStrategy.REPLACE)  
  Future<int> insert(T obj);  

  @Insert(onConflict: OnConflictStrategy.REPLACE)  
  Future<List<int>> insertAll(List<T> registers);  

  @Update()  
  Future<void> update(T obj);  

  @delete  
  Future<void> deleteEntity(T obj);  
}
```

And I used this class like that:

```
@dao
abstract class ContactDao extends BaseDao<Contact> {
  // ...
}
```

But the generated database file doesn't contain the base methods defined on BaseDao.
This kind of inheritance works fine on Android/Room. Can I do that with Floor? How?

If that is not possible, it would be a good feature to be added. 
Floor is amazing!!

joshuatam:
Can support similar implementations:

```dart
import 'package:floor/floor.dart';

abstract class BaseDao<T> {
  
  @insert
  Future<int> insertModel(T model);

  @update
  Future<void> updateModel(T model);
}

@dao
abstract class ItemDao extends BaseDao<Item> {
  // Other Queries...
}
```

Closes #258

### DIFF
--- a/floor_generator/lib/processor/dao_processor.dart
+++ b/floor_generator/lib/processor/dao_processor.dart
@@ -39,13 +39,9 @@ class DaoProcessor extends Processor<Dao> {
   @override
   Dao process() {
     final name = _classElement.displayName;
-    final methods = _classElement.methods;
-
-    dynamic _super = _classElement.supertype;
-    while (_super != null) {
-      methods.addAll(_super.methods);
-      _super = _super.superclass;
-    }
+    final supertypeMethods =
+        _classElement.allSupertypes.expand((type) => type.methods);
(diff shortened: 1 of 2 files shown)


vitusortner:
Thanks for your contribution! I've added tests and support for inheriting DAO methods that are defined in interfaces (done by using the `implements` keyword on a class) and mixins.

PR #289, opened 2020-03-18
Support inherited fields for entities and views

### LINKED ISSUE #188: database objects can't extend classes
This:
```
class A extends B {
  @PrimaryKey(autoGenerate: true)
  int id;
}

class B {
  int id;
```
will not work. Every time I query for A, its id will be null. Doesn't matter if it's a primary key or a manually set value. However, interestingly it does query correctly, just the returned result is overwritten somehow.

### LINKED ISSUE #270: Doesn't work with inheritance
All my entities has some common columns:

```dart
abstract class DataEntity {
  DataEntity({@required this.id, this.created, this.updated, this.deleted});

 @PrimaryKey()
  final String id;

  @ColumnInfo(nullable: false)
  final DateTime created;

  @ColumnInfo(nullable: false)
  final DateTime updated;

  @ColumnInfo(nullable: true)
  final DateTime deleted;
}
```

When inheriting from the above class, the code generator fails:

```dart
@Entity(tableName: "Settings", indices: [
  Index(value: ["created"]),
  Index(value: ["updated"]),
  Index(value: ["deleted"]),
])
class Setting extends DataEntity {
  Setting({
    @required String id,
    @required dynamic value,
    DateTime created,
    DateTime updated,
    DateTime deleted,
  })  : this.value = jsonEncode(value),
        super(
          id: id,
          created: created,
          updated: updated,
          deleted: deleted,
        );

  final String value;
}
```

Error:
```
[SEVERE] floor_generator:floor_generator on lib/data/data_context.dart (cached):
Error running FloorGenerator
There is no primary key defined on the entity Setting.
package:meucronogramacapilar/data/setting.dart:21:7
   ╷
21 │ class Setting extends DataEntity {
```

1) Floor should consider inherited classes (specially abstract ones)
2) It would make a lot more sense if there is a Index decorator for columns (so I don't need to declare indices from inherited classes every time)

vitusortner:
Closes #188 
Closes #270

### DIFF
--- a/floor_generator/lib/processor/queryable_processor.dart
+++ b/floor_generator/lib/processor/queryable_processor.dart
@@ -3,6 +3,7 @@ import 'package:analyzer/dart/element/type.dart';
 import 'package:floor_annotation/floor_annotation.dart' as annotations;
 import 'package:floor_generator/misc/annotations.dart';
 import 'package:floor_generator/misc/type_utils.dart';
+import 'package:floor_generator/processor/error/queryable_processor_error.dart';
 import 'package:floor_generator/processor/field_processor.dart';
 import 'package:floor_generator/processor/processor.dart';
 import 'package:floor_generator/value_object/field.dart';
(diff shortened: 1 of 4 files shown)


floor_generator/lib/processor/queryable_processor.dart
```diff
@@ -19,7 +19,12 @@ abstract class QueryableProcessor<T extends Queryable> extends Processor<T> {
   @nonNull
   @protected
   List<Field> getFields() {
-    return classElement.fields
+    final fields = [
+      ...classElement.fields,
+      ...classElement.allSupertypes.expand((type) => type.element.fields),
```
vitusortner: ❓ Are mixins included here?

PR #293, opened 2020-03-20
Support multiline database view queries

### LINKED ISSUE #292: DatabaseView generation error with multiline string
I'm using the 'develop' branch up to date to use the DatabaseView annotation (thanks by the way !).
It seems that the code generator doesn't work for multiline string requests.
```dart
@DatabaseView("""SELECT X.*
FROM X
GROUP BY X.a""")
```
It's quite surprising as multiline string are working for DAO.

vitusortner:
Closes #292

### DIFF
--- a/floor_generator/lib/misc/string_utils.dart
+++ b/floor_generator/lib/misc/string_utils.dart
@@ -3,4 +3,10 @@ extension StringUtils on String {
   String decapitalize() {
     return '${this[0].toLowerCase()}${substring(1)}';
   }
+
+  /// Flattens a multiline string into a single line string by concatenating
+  /// lines separated by a space and remove leading and trailing whitespace
+  String flatten() {
+    return replaceAll('\n', ' ').replaceAll(RegExp(r'[ ]{2,}'), ' ').trim();
+  }
 }
--- a/floor_generator/lib/misc/string_utils.dart
+++ b/floor_generator/lib/misc/string_utils.dart
@@ -4,8 +4,8 @@ extension StringUtils on String {
     return '${this[0].toLowerCase()}${substring(1)}';
   }
 
-  /// Flattens a multiline string into a single line string by concatenating
-  /// lines separated by a space and remove leading and trailing whitespace
+  /// Flattens a multiline string into a single-line string by concatenating
+  /// lines separated by a space and removes leading and trailing whitespace
   String flatten() {
     return replaceAll('\n', ' ').replaceAll(RegExp(r'[ ]{2,}'), ' ').trim();
   }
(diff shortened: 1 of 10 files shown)


vitusortner:
@mqus Would you mind reviewing?

mqus:
I'd be happy to review! Can you send me an invite? mqus at disroot dot org

floor_generator/lib/misc/string_utils.dart
```diff
@@ -3,4 +3,10 @@ extension StringUtils on String {
   String decapitalize() {
     return '${this[0].toLowerCase()}${substring(1)}';
   }
+
+  /// Flattens a multiline string into a single line string by concatenating
+  /// lines separated by a space and remove leading and trailing whitespace
+  String flatten() {
+    return replaceAll('\n', ' ').replaceAll(RegExp(r'[ ]{2,}'), ' ').trim();
```
mqus: I'd rather not mess with the query at all and just make the whole CREATE VIEW statement a multiline-string in database_writer.dart

Replacing newlines seems pretty safe, there could be rare cases where a newline is intentionally placed inside a string but I think not permitting that is not an issue, if it is properly documented. But eliminating multiple spaces without knowing the structure of the Query is dangerous. Making the writer output multiline strings seems safer and easier to me (I just tried a bit and it seemed to work)

mqus: In hindsight, this could also be possible for QueryMethods, but I haven't tried it.

floor_generator/lib/processor/view_processor.dart
```diff
@@ -43,10 +42,12 @@ class ViewProcessor extends QueryableProcessor<View> {
     final query = classElement
         .getAnnotation(annotations.DatabaseView)
         .getField(AnnotationField.VIEW_QUERY)
-        .toStringValue();
+        ?.toStringValue()
```
mqus: Is this `?` really neccessary here? Isn't the `query` field mandatory for `DatabaseView` annotations?

floor_generator/lib/misc/string_utils.dart
vitusortner: Good point. I was wondering if trimming leading and trailing whitespace would be required for SQLite to function as expected and turns out, trimming is not required. This leads to simpler code on our side! 🎉

floor_generator/lib/processor/view_processor.dart
vitusortner: Yes, `query` is mandatory. Unluckily, the code generator can still be run even if the code doesn't compile because the `query` is missing. In that case, the gernator produces a helpful error message.

PR #298, opened 2020-03-28
Use floor_annotation git ref during development

### LINKED ISSUE #288: Snapshot dependencies don't work
If I try to depend on floor like its written in the Snapshot section in the readme, it will throw the following error:
```
Error on line 21, column 11: Invalid description in the "floor_generator" pubspec on the "floor_annotation" dependency: "../floor_annotation/" is a relative path, but this isn't a local pubspec.
                                                                       
21      path: ../floor_annotation/                                     
              ^^^^^^^^^^^^^^^^^^^^                                     
```                        
Sadly, this does not seem to work, as others have already discovered:
https://stackoverflow.com/questions/54555612/how-to-organise-multi-package-flutter-project-and-use-it-as-dependency

It seems to be covered in this issue: https://github.com/dart-lang/pub/issues/449

Personally, I'll probably work around this by using git submodules in my project.

vitusortner:
This allows using snapshot versions of the library. We can't use local paths for snapshot versions as Pub disallows relative paths. I've decided for this solution as we won't change `floor_annotation` that frequently.

Closes #288

### DIFF
--- a/example/lib/database.g.dart
+++ b/example/lib/database.g.dart
@@ -74,7 +74,7 @@ class _$FlutterDatabase extends FlutterDatabase {
         await callback?.onOpen?.call(database);
       },
       onUpgrade: (database, startVersion, endVersion) async {
-        MigrationAdapter.runMigrations(
+        await MigrationAdapter.runMigrations(
             database, startVersion, endVersion, migrations);
 
         await callback?.onUpgrade?.call(database, startVersion, endVersion);
--- a/floor/pubspec.yaml
+++ b/floor/pubspec.yaml
@@ -13,7 +13,10 @@ dependencies:
   path: ^1.6.4
   sqflite: ^1.2.0
   floor_annotation:
-    path: ../floor_annotation/
+    git:
+      url: git://github.com/vitusortner/floor.git
+      ref: snapshot-versions
+      path: floor_annotation
   meta: ^1.1.8
   flutter:
     sdk: flutter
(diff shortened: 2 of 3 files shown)


vitusortner:
I just realized that with these changes, we won't be able to use newly added annotations until they're available on `develop`. This would mean that we'll aways be required to push the annotations to `develop` before, for instance, our CI can use them.

Do you have another idea how to solve the `"../floor_annotation/" is a relative path, but this isn't a local pubspec.` problem? @mqus

mqus:
Not really... 
I see at least three options, all have their downsides:

1. We could split PRs which change annotations in two subsequent PRs, in which one introduces the change to floor_annotation (and can't really be tested) and the following one updates the generator/floor library... this could happen concurrently by referencing the annotation-PR branch while developing in the other branch and switching that after the annotation PR was merged....
If i understand it right then this is what you already proposed.

2. Any PR changing the annotations has to change all references in the pubspec to its PR branch, which will persist until the release or another commit updates the reference again.

3. We keep the current relative annotations and release often. Everyone who still wants to use the develop branch has to use a local copy by either using git submodules or just extracting a snapshot into their sources.

The downsides of 3 were already discussed, 1 and 2 are pretty complex and we would have to apply them to all drive-by PRs. Otoh, changes to floor_annotation seem to be fairly rare, so this should only be necessary for a handful of PRs.

Personally, I'd like to avoid complexities in the PR process so I'm advocating for 3. But I don't have a strong opinion on this, so this would be up to you ;-)

vitusortner:
I'll close this PR for now and [remove the snapshot section from the README](https://github.com/vitusortner/floor/pull/301).

PR #305, opened 2020-04-06
Update README.md

ManoyloK:
Fixed return type


vitusortner:
Thanks, good find! There is a copy of the same README located [here](https://github.com/vitusortner/floor/blob/develop/floor/README.md). Please also update the file there 🚀

PR #315, opened 2020-04-17
add isRaw to Query annotation

joshuatam:
Can achieve raw query like this:

```dart
@Query('SELECT * FROM Person WHERE :condition', isRaw: true)
Future<Person> findByCondition(String condition);
```
and condition like:
```dart
String condition = "isDeleted = 0 AND name = '$name'";
```


mqus:
Are you finished already and want a review or is this WIP? If its a work in progress, please convert this PR into a draft for now(The option should be somewhere on the right). 
If not, please add some test(s) and I will review.

joshuatam:
@mqus Any better ideas to accomplish this feature?
This feature is useful if I need to build the query directly and dynamically.

mqus:
I think the architecture of floor (or room for that matter) is not well-suited for dynamic querying or query-building, but your approach can definitely work and be a middle ground/workaround for that. 

We could try to move to a more complex and powerful [RawQuery](https://developer.android.com/reference/androidx/room/RawQuery) like room has in the future but for now I think this is sufficient.

What are your thoughts, @vitusortner ?

TalebRafiepour:
I need this feature now, How can use your commit in my project?

PR #318, opened 2020-04-21
Add support for type converters

### LINKED ISSUE #165: Add support for type converters
I read a mention of this in #119. I think it's a great idea.

I will be digging the code and seeing if I can help out, but thought I'd open this issue anyway.

Also, thanks a lot for the great library!

### LINKED ISSUE #313: Can't primaryKey be String type?
When I create a entity like this: 
```
class User {
  @PrimaryKey()
  final String id;
}
```

After `flutter packages pub run build_runner build`, the dao mapper of generated in `*.g.dart` file is :
```
id: row['id'] as double,
```

I need to change it manually every build as:

```
id: row['id'] as String,
```
Otherwise it won't work when operate db :

```
type 'String' is not a subtype of type 'double' in type cast
```

Is this a bug？

Version：

```
dependencies:
  floor: ^0.12.0

dev_dependencies:
  floor_generator: ^0.12.0
  build_runner: ^1.7.3
```

vitusortner:
### Public API

First, some words about the public API of type converters. I wanted to make type converters as safe as possible at compile time. Thereby, I added an abstract `TypeConverter` class that required two type parameters. The first is the encoded type that is used outside of the database scope. The second type is the decoded database type. The abstract class, furthermore, comes with the two functions `encode()` and `decode()` which have to be overridden by the implementation classes. This way, we can be sure that there always is functionality provided for conversion in both directions. Room is missing this extra security as they use [plain functions as type converters](https://developer.android.com/training/data-storage/room/referencing-data#type-converters).

A simple type converter that converts from `DateTime` to `int` and vice versa is shown in the following.
```dart
class DateTimeConverter extends TypeConverter<DateTime, int> {
  @override
  DateTime decode(int databaseValue) {
    return DateTime.fromMillisecondsSinceEpoch(databaseValue);
  }

  @override
  int encode(DateTime value) {
    return value.millisecondsSinceEpoch;
  }
}
```

Such a class can then be applied to the database with different scope restrictions by using the `@TypeConverters` annotation. These scopes are:
1. databases
1. DAOs
1. entities/views
1. entity/view fields
1. DAO methods
1. DAO method parameters

### Implementation

Floor generates and uses type converter as described in the following.

**Processors**
1. Collect type converters with the database scope and pass these to all processors
1. Collect type converters with the DAO scope and pass these to the specific DAO processor
1. This basically happens for each scope level and thereby, the processors only get the relevant type converters

**Writers**
1. Instantiate/write an object of each type converter class it can find (it has to be added with `@TypeConverters`)
1. Use the generated type converters throughout the generated code when needed

This still needs work:
- [ ] tests (validate type converter scopes/order and correct usage)

Closes #165
Closes #313 as we won't fall back to `double` when the type of a field is unknown


floor_generator/lib/misc/extension/type_converters_extension.dart
```diff
@@ -0,0 +1,32 @@
...
+  TypeConverter get closestOrNull {
+    return sortedByDescending((typeConverter) => typeConverter.scope.index)
+        .firstOrNull;
+  }
+
+  @nullable
+  TypeConverter getClosestOrNull(DartType dartType) {
+    return sortedByDescending((typeConverter) => typeConverter.scope.index)
+        .firstOrNullWhere(
+            (typeConverter) => typeConverter.fieldType == dartType);
```
vitusortner, line 18: ❓Does `is` work here?
```suggestion
            (typeConverter) => typeConverter.fieldType is dartType);
```

floor/test/integration/type_converter/order_database.dart
```diff
@@ -0,0 +1,18 @@
+import 'dart:async';
+
+import 'package:floor/floor.dart';
+import 'package:floor_annotation/floor_annotation.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:sqflite/sqflite.dart' as sqflite;
+
+import 'order.dart';
+import 'order_dao.dart';
+import 'type_converter.dart';
```
vitusortner, line 10: Point out that the type converter import can't be added automatically and thus has to be mentioned in the database definition.

floor/test/integration/type_converter/order_dao.dart
```diff
@@ -0,0 +1,24 @@
...
+
+  @Query('SELECT * FROM `Order` WHERE id = :id')
+  Future<Order> findOrderById(int id);
+
+  @Query('SELECT * FROM `Order` WHERE date = :date')
+  Future<List<Order>> findOrdersByDate(DateTime date);
+
+  @Query('SELECT * FROM `Order`')
+  Future<List<Order>> findAllOrders();
+}
```
vitusortner, line 24: Are all of these methods used? If not, delete.

floor_annotation/lib/floor_annotation.dart
```diff
@@ -14,4 +14,6 @@ export 'src/on_conflict_strategy.dart';
 export 'src/primary_key.dart';
 export 'src/query.dart';
 export 'src/transaction.dart';
+export 'src/type_converter.dart';
```
vitusortner, line 17: Create an issue to move all non-Flutter code into the floor package (annotations included) and move Flutter dependent code into the floor_flutter package.

floor_generator/lib/generator.dart
```diff
@@ -23,18 +24,31 @@ class FloorGenerator extends GeneratorForAnnotation<annotations.Database> {
...
-    final daoClasses = daoGetters.map((daoGetter) => DaoWriter(
-            daoGetter.dao, database.streamEntities, database.hasViewStreams)
-        .write());
-
-    final library = Library((builder) => builder
-      ..body.add(FloorWriter(database.name).write())
-      ..body.add(DatabaseBuilderWriter(database.name).write())
-      ..body.add(databaseClass)
-      ..body.addAll(daoClasses));
+    final daoClasses = daos.map((dao) => DaoWriter(
```
vitusortner, line 30: Merge assignments.

floor_generator/lib/writer/query_method_writer.dart
```diff
@@ -76,36 +68,40 @@ class QueryMethodWriter implements Writer {
...
-            return '''final valueList$index = ${parameter.displayName}.map((value) => "'\$value'").join(', ');''';
-          } else {
-            return null;
-          }
-        })
-        .where((string) => string != null)
-        .toList();
+        .where((parameter) => parameter.type.isDartCoreList)
+        .mapIndexed((index, parameter) {
+      // TODO #165 what about type converters that map between e.g. string and list?
```
mqus, line 74: I think we can leave that out for now and create a new issue for that to discuss the possibilities. Making a wrapper method which converts this is a simple thing for a workaround for now.

mqus:
Regarding the differences to Room: room uses static methods (which can't be used in inheritance structures), which have the bonus that no objects have to be instantiated and the compiler could inline the calls when compiling the generated code. Also: No TypeConverter constructors have to be created and managed as the static methods simply can be called from anywhere. Checking that both conversion methods exist is 1) maybe not always necessary (e.g. if a custom type is only used as a parameter in a dao method) and 2) feasible with the static analysis we already do, we don't really need the type system to help us here.

I don't know if we can also add the room way to floor in the future and simply allow both... But that would also make the type converter plumbing more complex.

floor_generator/lib/processor/queryable_processor.dart
```diff
@@ -62,40 +77,52 @@ abstract class QueryableProcessor<T extends Queryable> extends Processor<T> {
...
-      return '$parameterValue as Uint8List';
+    } else if (dartType.isDartCoreString) {
+      return '$this as String';
+    } else if (dartType.isDartCoreInt) {
+      return '$this as int';
+    } else if (dartType.isUint8List) {
+      return '$this as Uint8List';
     } else {
-      return '$parameterValue as double'; // must be double
+      return '$this as double'; // must be double
```
mqus, line 125: I would rather have a `else if` here and an error if it does not match anything, see #313

floor_generator/lib/processor/type_converter_processor.dart
```diff
@@ -0,0 +1,39 @@
...
+  TypeConverter process() {
+    final typeArguments = _classElement.supertype.typeArguments;
+    final fieldType = typeArguments[0];
+    final databaseType = typeArguments[1];
+
+    if (!databaseType.isDefaultSqlType) {
+      throw InvalidGenerationSourceError(
+        'Type converters have to convert to a database-compatible type.',
+        todo:
+            'Make the class convert to either int, double, String, bool or Uint8List.',
```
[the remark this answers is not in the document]
mqus: Update: I think we should keep bool there, just because it is also a special type in sqlparser and it could come in handy in the future. But I'm still interested in your thoughts :)

floor_annotation/lib/floor_annotation.dart
vitusortner: Let's see what we can do to not introduce a circular dependency.

My thoughts on re-structuring the code were:
- floor - holds annotations and needed building blocks that users either use directly or the generated code uses (e.g. `InsertionAdapter`)
  - no dependencies on other floor packages
- floor_flutter - holds everything with a direct dependency to Flutter (e.g. sqflite)
  - depends on floor
  - has dev dependency on floor_generator
- floor_generator - holds the generator that users will include as a dev dependency
  - depends on floor

floor/README.md
```diff
@@ -517,6 +520,64 @@ await personDao.insertItem(person);
...
+#### Be aware that type converters can be applied to:
+1. databases
+1. DAOs
+1. entities/views
+1. entity/view fields
+1. DAO methods
+1. DAO method parameters
+
+The type converter is added to the scope of the element so if you put it on a class, all methods/fields in that class will be able to use the converters.
+
```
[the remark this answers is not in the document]
vitusortner: Great point! Will do 👍

floor_generator/lib/processor/type_converter_processor.dart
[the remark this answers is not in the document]
vitusortner: Let's leave our bool handling as it is 👍 
About introducing default type converters: I'm not a big fan of that as it hides functionality which might lead to unexpected behavior. On the other hand, we basically have a type converter built into the library as we're converting bools everywhere 🙃

floor_generator/lib/misc/type_utils.dart
```diff
@@ -8,7 +8,7 @@ import 'package:source_gen/source_gen.dart';
 
 extension SupportedTypeChecker on DartType {
   @nonNull
-  bool get isSupported {
+  bool get isDefaultSqlType {
     return TypeChecker.any([
       _stringTypeChecker,
       _boolTypeChecker,
```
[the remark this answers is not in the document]
vitusortner: Good point!

vitusortner:
> Regarding the differences to Room: room uses static methods (which can't be used in inheritance structures), which have the bonus that no objects have to be instantiated and the compiler could inline the calls when compiling the generated code. Also: No TypeConverter constructors have to be created and managed as the static methods simply can be called from anywhere. Checking that both conversion methods exist is 1) maybe not always necessary (e.g. if a custom type is only used as a parameter in a dao method) and 2) feasible with the static analysis we already do, we don't really need the type system to help us here.
> 
> I don't know if we can also add the room way to floor in the future and simply allow both... But that would also make the type converter plumbing more complex.

These are some very valid points you're making here. I gave this some thought which makes me tend more towards using static functions instead of the implementation of the abstract `TypeConverter` class. Especially the fact that both conversion directions might not always be needed, makes me want to change things.
In one of the early drafts of the type converter implementation, I was using top-level functions as converters. These, unluckily, would come with the downside that users would need to add each converter explicitly as the chance to group them would be lost (e.g. `@TypeConverters(fromAToB, fromBToA)` instead of `@TypeConverters(AToBConverter)`). This gives a good reason to group the converter functions in a class or abstract class. On the other hand, there is an analyzer rule that warns about classes with only static functions and suggests to use top-level functions instead (`analysis_options.dart - avoid_classes_with_only_static_members`).

This is what type converter definitions from Room would look like in Dart.
```dart
class AToBConverter {
  static B convertAToB(A a) => a.toB();
  static A convertBToA(B b) => b.toA();
}
```

Do you have a suggestion on how to overcome the mentioned concerns?

avbk:
Hey @vitusortner and @mqus, first of all, thanks for your awesome work! I'd like to put in my two cents as well and I hope you don't mind :slightly_smiling_face: 

> Regarding the differences to Room: room uses static methods (which can't be used in inheritance structures), which have the bonus that no objects have to be instantiated and the compiler could inline the calls when compiling the generated code. 

Actually I think the performance impact is more or less insignificant compared to having sql queries being executed and results sent via methodchannels. But how about providing a `const` constructor for `TypeConverter`. The compiler should be able to optimize such classes and methods, I guess? It would though mean more boilerplate code for the developer using this library, since it would be necessary to provide an `const` constructor for the inheriting class as well. But one could omit it, if he doesn't care about performance...

> Also: No TypeConverter constructors have to be created and managed as the static methods simply can be called from anywhere. 

As far as I understand the converters are held as file-private fields in the generated database file. So I guess the effect of creating the converters is negligible.

> Checking that both conversion methods exist is 1) maybe not always necessary (e.g. if a custom type is only used as a parameter in a dao method) and 2) feasible with the static analysis we already do, we don't really need the type system to help us here.

Honestly I haven't had a situation yet, where I only needed decoding or encoding but not both. Still you have a point and maybe you could introduce a `DecodingTypeConverter<T, S>` as well as a `EncodingTypeConverter<T, S>` and let `TypeConverter<T, S> implements DecodingTypeConverter<T, S>, EncodingTypeConverter<T, S>`. This would allow one to have only one part of conversion, if he needs to, but by default he gets a nice compile error, if he uses the TypeConverter incorrectly.

proninyaroslav:
@vitusortner
Hi. Do you have an approximate release date for this feature?

fnicastri:
@vitusortner @mqus 
Just a thought: what about making the converters compatible with json_serializable?
It's just an idea, I did't looked at the code...

mqus:
@fnicastri what do you mean with "compatible"? What interactions would you like to do that are not possible today, could you give examples? I have to confess that I'm not that knowledgeable about dart yet and while I have worked with `json_serializable` (or something similar, don't remember), I can't remember how to use it.

fnicastri:
@mqus 
To be honest I had not time to dig in it, sorry, I was planning to do that just after my comment but then work work work...
The reasoning behind my comment is:
probably we use the two libs together a lot to ingest json, save them in the db, use the data and send the data back( at least this is what I do).
If we can use the same converters for both we will save a lot of code. I think especially about complex objects or fields that need some operations before the serialization.

This can be, obviously, a useless idea or something that need too much changes to be feasible.

Best,
Frank

mqus:
@fnicastri Ah right I get it now, I probably misunderstood you before, sorry. This is a good idea, although I haven't looked at how the type converters for `json_serializable` are implemented yet and if this would be feasible.

avbk:
@mqus, @fnicastri 

Hey guys :slightly_smiling_face: 

I worked with `json_serializable` a couple of times and I'm not sure if the converters of `json_serializable` fit with the type converters of this PR.  There is a small but important difference between the two converter interfaces, which makes them very similar but not identical: 
* the converter methods in this PR convert between a single database primitive and a type `T`
* the converter methods of `json_serializable` convert between a single json literal and a type `T`

For simple cases like converting ISO8601 formatted strings into `DateTime` and vice versa the methods would be identical. But a json literal does not need to be a `String` or an `int`, it can also be a `Map<String, dynamic>`. Therefore in `json_serializable` you can use the following valid code, which would not work as a type converter for floor (actually it resembles the idea of `@Embedded` in a way):
```dart

@JsonSerializable()
class PointOfInterest {
  String title;

  @JsonKey(fromJson: _latLngFromJson, toJson: _latLngToJson)
  LatLng position;

  PointOfInterest(this.title, this.position);

  factory Poi.fromJson(Map<String, dynamic> json) => _$PointOfInterestFromJson(json);
  Map<String, dynamic> toJson() => _$PointOfInterestToJson(this);
}

Map<String, dynamic> _latLngToJson(LatLng latlng) =>
    {"latitude": latlng.latitude, "longitude": latlng.longitude};

LatLng _latLngFromJson(Map<String, dynamic> json) =>
    LatLng(json["latitude"] as double, json["longitude"] as double);
```
And the corresponding json would be:
```json
{
  "title": "A lovely spot in the ocean",
  "position": {
    "latitude": 0.0,
    "longitude": 0.0
  }
}
```

Apart from that I think you can support both converters with just a little boilerplate by delegating the two calls:
```dart
enum Taste { delicious, disgusting }

@JsonSerializable()
class RecipeDTO {
  String title;
  @JsonKey(fromJson: _stringToTaste, toJson: _tasteToString)
  Taste taste;

  // ...
}

@Entity
@TypeConverters([_TasteConverter])
class Recipe {
  String title;
  Taste taste;

  // ...
}

String _tasteToString(Taste taste) => /* ... */
Taste _stringToTaste(String json) => /* ... */

class _TasteConverter extends TypeConverter<Taste, String> {
  @override
  Taste decode(String databaseValue) => _stringToTaste(databaseValue);

  @override
  String encode(Taste value) => _tasteToString(value); 
}
```

I guess a real benefit would emerge (assuming you want to save the DTO directly in the database), if the Entity would support `json_serializable` methods (just like [freezed](https://pub.dev/packages/freezed#fromjsontojson) is doing). But supporting this comes with some problems, e.g. json can be nested while entities can only be flat. 

Nevertheless I find it an interesting idea to support it for smaller use cases, but I think it should be addressed in a separate issue and not within this PR... 

But that's just my humble opinion and I'm curios about what you guys think :wink:

mqus:
Yeah I looked at it briefly yesterday and came to pretty much the same conclusions. The mismatch between json types/structure and sqlite types/structure is a bit too much. Delegates should be fine. Regarding your example: if you don't take the (non-)flatness into account then you could also merge `RecipeDTO` and `Recipe` as floor doesn't mind the annotations it doesn't understand (and I'm pretty sure JsonSerializable handles it the same)

fnicastri:
@mqus 
Thanks for your time, it was just an idea.
@avbk 
Yu can definitely merge `Recipe` and `RecipeDTO`, the only downside is when you have a lot of annotations, it starts to feel messy

FingerArt:
@vitusortner
Hi. Do you have an approximate release date for this feature?

FingerArt:
@vitusortner
Hi. Do you have an approximate release date for this feature?

FlashLight13:
@vitusortner Hi! Any updates on this? 🙂

vitusortner:
I finally have some more time to look into this feature again. I'll keep you updated!

floor_generator/lib/writer/query_method_writer.dart
vitusortner: Will be solved in #403.

floor_generator/lib/misc/type_utils.dart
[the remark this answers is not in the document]
vitusortner: It appears to me that we don't ever need the flattened type here. Our tests also run as expected which makes me want to get rid of the flattening logic 🤔

PR #320, opened 2020-05-01
Implement simple Streams on DatabaseViews, fix multi-dao changelistener

mqus:
Like I explained on the slack, here is a prototype for Queries with the result `Stream<X>` where X is a `@DatabaseView`.

As it won't detect the dependencies of each `DatabaseView`, it simply updates the `Stream`s of all views if *any* entity was updated. If a library user is frequently changing his entities while having one particularly intensive view (like a join) as a `Stream`, it will get quite resource-heavy even if the entity and the `DatabaseView` of the `Stream` are not related at all. But its better than nothing.

Of course, I think we should strive to make the `Stream`s of `DatabaseView`s better by making them aware of the entities they depend on but as mentioned previously this can get very complex.

I also fixed a related issue where `Stream`s of `Entity` E in DAO B were only updated by insert/update/delete methods in DAO C if there was another query of a Stream of E in C because Streams were only tracked separately for each DAO even if the `changeListener` was database-global.

## TODO:
- [x] remove error message for stream queries on views (it doesn't get thrown anymore but it's still there)
- [x] implement tests
    - [x] implement Tests (only adapted existing ones so far) for the Streams on Views
    - [x] implement Test for the fix
- [x] implement integration tests
    - [x] implement Tests for the Streams on Views
    - [x] implement Regression Test for the issue
- [x] add documentation and mention that it's expensive

I'll finish this up if you agree on the concept, @vitusortner .

### DIFF
--- a/floor/test/adapter/query_adapter_test.dart
+++ b/floor/test/adapter/query_adapter_test.dart
@@ -259,6 +259,30 @@ void main() {
         ]),
       );
     });
-    //TODO test view updates every time
+
+    test('query stream from view with same and different triggering entity',
+        () async {
+      final person = Person(1, 'Frank');
+      final person2 = Person(2, 'Peter');
+      final queryResult = Future(() => [
+            <String, dynamic>{'id': person.id, 'name': person.name},
+            <String, dynamic>{'id': person2.id, 'name': person2.name},
+          ]);
+      when(mockDatabaseExecutor.rawQuery(sql)).thenAnswer((_) => queryResult);
+
+      final actual = underTest.queryListStream(sql,
+          queryableName: entityName, isView: true, mapper: mapper);
+      expect(
+        actual,
+        emitsInOrder(<List<Person>>[
+          <Person>[person, person2],
+          <Person>[person, person2],
+          <Person>[person, person2]
+        ]),
+      );
+
+      streamController.add(entityName);
+      streamController.add('otherEntity');
+    });
   });
 }
--- a/floor/test/adapter/query_adapter_test.dart
+++ b/floor/test/adapter/query_adapter_test.dart
@@ -162,7 +162,7 @@ void main() {
       when(mockDatabaseExecutor.rawQuery(sql)).thenAnswer((_) => queryResult);
 
       final actual = underTest.queryStream(sql,
-          tableName: entityName, isView: false, mapper: mapper);
+          queryableName: entityName, isView: false, mapper: mapper);
 
       expect(actual, emits(person));
     });
@@ -179,7 +179,7 @@ void main() {
       final actual = underTest.queryStream(
         sql,
         arguments: arguments,
-        tableName: entityName,
+        queryableName: entityName,
         isView: false,
         mapper: mapper,
       );
@@ -195,7 +195,7 @@ void main() {
       when(mockDatabaseExecutor.rawQuery(sql)).thenAnswer((_) => queryResult);
 
       final actual = underTest.queryStream(sql,
-          tableName: entityName, isView: false, mapper: mapper);
+          queryableName: entityName, isView: false, mapper: mapper);
       streamController.add(entityName);
 
       expect(actual, emitsInOrder(<Person>[person, person]));
@@ -211,7 +211,7 @@ void main() {
       when(mockDatabaseExecutor.rawQuery(sql)).thenAnswer((_) => queryResult);
 
       final actual = underTest.queryListStream(sql,
-          tableName: entityName, isView: false, mapper: mapper);
+          queryableName: entityName, isView: false, mapper: mapper);
 
       expect(actual, emits([person, person2]));
     });
@@ -230,7 +230,7 @@ void main() {
       final actual = underTest.queryListStream(
         sql,
         arguments: arguments,
-        tableName: entityName,
+        queryableName: entityName,
         isView: false,
         mapper: mapper,
       );
@@ -248,7 +248,7 @@ void main() {
       when(mockDatabaseExecutor.rawQuery(sql)).thenAnswer((_) => queryResult);
 
       final actual = underTest.queryListStream(sql,
-          tableName: entityName, isView: false, mapper: mapper);
+          queryableName: entityName, isView: false, mapper: mapper);
       streamController.add(entityName);
 
       expect(
@@ -259,5 +259,6 @@ void main() {
         ]),
       );
     });
+    //TODO test view updates every time
   });
 }
--- a/floor/test/adapter/query_adapter_test.dart
+++ b/floor/test/adapter/query_adapter_test.dart
@@ -161,8 +161,8 @@ void main() {
           ]);
       when(mockDatabaseExecutor.rawQuery(sql)).thenAnswer((_) => queryResult);
 
-      final actual =
-          underTest.queryStream(sql, tableName: entityName, mapper: mapper);
+      final actual = underTest.queryStream(sql,
+          tableName: entityName, isView: false, mapper: mapper);
 
       expect(actual, emits(person));
     });
@@ -180,6 +180,7 @@ void main() {
         sql,
         arguments: arguments,
         tableName: entityName,
+        isView: false,
         mapper: mapper,
       );
 
@@ -193,8 +194,8 @@ void main() {
           ]);
       when(mockDatabaseExecutor.rawQuery(sql)).thenAnswer((_) => queryResult);
 
-      final actual =
-          underTest.queryStream(sql, tableName: entityName, mapper: mapper);
+      final actual = underTest.queryStream(sql,
+          tableName: entityName, isView: false, mapper: mapper);
       streamController.add(entityName);
 
       expect(actual, emitsInOrder(<Person>[person, person]));
@@ -209,8 +210,8 @@ void main() {
           ]);
       when(mockDatabaseExecutor.rawQuery(sql)).thenAnswer((_) => queryResult);
 
-      final actual =
-          underTest.queryListStream(sql, tableName: entityName, mapper: mapper);
+      final actual = underTest.queryListStream(sql,
+          tableName: entityName, isView: false, mapper: mapper);
 
       expect(actual, emits([person, person2]));
     });
@@ -230,6 +231,7 @@ void main() {
         sql,
         arguments: arguments,
         tableName: entityName,
+        isView: false,
         mapper: mapper,
       );
 
@@ -245,8 +247,8 @@ void main() {
           ]);
       when(mockDatabaseExecutor.rawQuery(sql)).thenAnswer((_) => queryResult);
 
-      final actual =
-          underTest.queryListStream(sql, tableName: entityName, mapper: mapper);
+      final actual = underTest.queryListStream(sql,
+          tableName: entityName, isView: false, mapper: mapper);
       streamController.add(entityName);
 
       expect(
--- a/floor/test/integration/stream_query_test.dart
+++ b/floor/test/integration/stream_query_test.dart
@@ -1,7 +1,6 @@
 import 'package:flutter_test/flutter_test.dart';
 
 import '../test_util/extensions.dart';
-import '../test_util/separate_queries.dart';
 import 'dao/person_dao.dart';
 import 'database.dart';
 import 'model/dog.dart';
@@ -143,7 +142,7 @@ void main() {
             [dog1], // after inserting dog1
             [dog1], // after inserting dog2
             //[], // after removing person1. Does not work because
-            // ForeignKey-relations are not considered. (TODO)
+            // ForeignKey-relations are not considered yet (#321)
           ]));
 
       await personDao.insertPerson(person1);
@@ -155,7 +154,7 @@ void main() {
 
       // avoid that delete happens before the re-execution of
       // the select query for the stream
-      await simulateDoingSomethingElse();
+      await Future<void>.delayed(const Duration(milliseconds: 100));
 
       await database.personDao.deletePerson(person1);
     });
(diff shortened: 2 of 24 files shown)


floor_generator/lib/processor/dao_processor.dart
```diff
@@ -56,6 +56,7 @@ class DaoProcessor extends Processor<Dao> {
     final transactionMethods = _getTransactionMethods(methods);
 
     final streamEntities = _getStreamEntities(queryMethods);
+    final streamViews = _getStreamViews(queryMethods);
```
vitusortner, line 59: 💡This functionality could be simplified and inlined.

```suggestion
    final streamQueryables = queryMethods
        .where((method) => method.returnsStream)
        .map((method) => method.queryable);

    final streamEntities = streamQueryables.whereType<Entity>().toList();
    final streamViews = streamQueryables.whereType<View>().toList();
```

floor_generator/lib/value_object/dao.dart
```diff
@@ -25,6 +27,7 @@ class Dao {
     this.deletionMethods,
     this.transactionMethods,
     this.streamEntities,
+    this.streamViews,
```
vitusortner, line 30: ⚠️Please update equals, hashCode and toString

floor_generator/lib/writer/query_method_writer.dart
```diff
@@ -137,11 +138,15 @@ class QueryMethodWriter implements Writer {
     @nullable final String arguments,
     @nonNull final String mapper,
   ) {
-    final entityName = _queryMethod.queryable.name;
-
+    final tableName = _queryMethod.queryable.name;
```
vitusortner, line 141: 💡Queryable names don't exclusively define table names but also views. What do you think about the following?

```suggestion
    final queryableName = _queryMethod.queryable.name;
```

floor_generator/lib/writer/query_method_writer.dart
```diff
@@ -137,11 +138,15 @@ class QueryMethodWriter implements Writer {
...
+    final isView =
+        _queryMethod.queryable.runtimeType == View ? 'true' : 'false';
```
vitusortner, line 143: 💡Dart comes with `is` as a language feature. You can simplify this code with the following.
```suggestion
    final isView = _queryMethod.queryable is View ? 'true' : 'false';
```

❓Did you look into `bool`'s `toString()`? Is there the need for explicitly defining 'true' and 'false' instead of just invoking `isView.toString()`? Do you think it's better to be explicit here?

floor_generator/lib/generator.dart
```diff
@@ -26,8 +26,9 @@ class FloorGenerator extends GeneratorForAnnotation<annotations.Database> {
     final daoGetters = database.daoGetters;
 
     final databaseClass = DatabaseWriter(database).write();
-    final daoClasses =
-        daoGetters.map((daoGetter) => DaoWriter(daoGetter.dao).write());
+    final daoClasses = daoGetters.map((daoGetter) => DaoWriter(
+            daoGetter.dao, database.entityStreams, database.hasViewStreams)
```
mqus, line 30: the fix part 2: All entitystreams and viewstream indicators of the database(=all DAOs) are given to the DaoWriter, instead of the DaoWriter only considering the entitystreams and viewstreams of its own Dao

floor_generator/lib/value_object/database.dart
```diff
@@ -20,7 +22,9 @@ class Database {
     this.views,
     this.daoGetters,
     this.version,
-  );
+  )   : entityStreams =
+            daoGetters.expand((dg) => dg.dao.streamEntities).toSet(),
+        hasViewStreams = daoGetters.any((dg) => dg.dao.streamViews.isNotEmpty);
```
mqus, line 27: the fix part 1: Accumulate the streams of all Daos in the database

floor_generator/lib/writer/query_method_writer.dart
mqus: no, toString seems good here. I'll change the code.

floor_generator/lib/value_object/dao.dart
mqus: I'll also use `ListEquality` for this and the other `List`s instead of the current ` == ` comparison.

cassioseffrin:
Do you have an estimative when streams on views will be released?

mqus:
Not exactly but maybe soon. There are just some tests left to do which I will try to tackle tomorrow. vitusortner will review again afterwards, although there shouldn't be major issues left, so this would then be ready to be merged. I don't know if there are plans for a new release.

But as mentioned, this is just a (possibly runtime-intensive)stopgap solution until I'm finished with integrating the sqlparser, which will do query-analysis to be more efficient when updating the Stream.
But I really can't give an ETA for that and help is appreciated.

cassioseffrin:
Query-analysis will be awesome! I am waiting for it!

mqus:
@vitusortner I noticed that we actually don't test streams quite the right way. Floor's Streams will re-execute the query as soon someone will start to Listen (which is right this way), but that also means that each 
```dart
expect(actual, emits([anything]));
```
in the tests just starts listening and reexecutes the query, not actually looking if an event was fired independently. Previously fired events will be discarded if no-one is actually listening. This means that our Stream tests are just checking if the insert/update/delete query got executed, not if the stream fired.

For the new tests added here I'm using 
```dart
expect(actual, emitsInOrder([firstResult,secondResult]));
```
at the start of the test, which will keep the stream open and actually checks the stream output. 

When doing this I noticed that some queries will be executed together before the re-execution of the Stream Query fires, meaning that some events get "lost" and other events will then get a double update. 

For that I introduced a new helper function that will introduce artificial wait times to let floor reexecute the SELECT query before the test starts the next database action. I'll push that soon.

mqus:
The last commit is not really necessary... you decide if you want it :)

floor_generator/test/processor/query_method_processor_test.dart
```diff
@@ -257,28 +253,30 @@ void main() {
...
+//      @Query('SELECT * FROM Name')
+//      Stream<List<Name>> allNamesAsStream();
+//    ''');
+//
+//      final actual =
+//          () => QueryMethodProcessor(methodElement, entities, views).process();
+//
+//      final error = QueryMethodProcessorError(methodElement).viewNotStreamable;
+//      expect(actual, throwsInvalidGenerationSourceError(error));
+//    });
```
vitusortner, line 279: This can be deleted.

floor_generator/test/writer/dao_writer_test.dart
```diff
@@ -177,6 +182,220 @@ void main() {
...
+    ''');
+    // simulate DB is aware of another streamed Entity and no View
+    final otherEntity = Entity(
+      null, //ClassElement classElement,
+      'Dog', //String name,
+      [], //List<Field> fields,
+      null, // this.primaryKey,
+      [], // this.foreignKeys,
+      [], // this.indices,
+      '', // String constructor
```
vitusortner, line 278:
```suggestion
      null, // classElement,
      'Dog', // name,
      [], // fields,
      null, // primaryKey,
      [], // foreignKeys,
      [], // indices,
      '', // constructor
```

floor_generator/lib/writer/query_method_writer.dart
```diff
@@ -141,11 +142,14 @@ class QueryMethodWriter implements Writer {
     @nullable final String arguments,
     @nonNull final String mapper,
   ) {
-    final entityName = _queryMethod.queryable.name;
-
+    final queryableName = _queryMethod.queryable.name;
+    final isView = (_queryMethod.queryable is View).toString();
```
vitusortner, line 146: 💡The explicit `toString()` shouldn't be required as the parameter is being used with string interpolation which automatically invokes `toString()` on the boolean field.
```suggestion
    final isView = _queryMethod.queryable is View;
```

floor_generator/lib/processor/dao_processor.dart
```diff
@@ -66,14 +70,15 @@ class DaoProcessor extends Processor<Dao> {
...
+      streamViews,
     );
   }
 
   List<QueryMethod> _getQueryMethods(final List<MethodElement> methods) {
     return methods
         .where((method) => method.hasAnnotation(annotations.Query))
         .map((method) =>
-            QueryMethodProcessor(method, _entities, _views).process())
+            QueryMethodProcessor(method, [..._entities, ..._views]).process())
```
vitusortner, line 81: ❤️Much cleaner, thanks!

floor/test/integration/view/view_test.dart
```diff
@@ -85,6 +104,48 @@ void main() {
...
+                Name('Leonhard'),
+                Name('Romeo')
+              ],
+              [Name('Frank')], // after removing Person (and associated Dog)
+            ]));
+
+        final persons = [Person(1, 'Leo'), Person(2, 'Frank')];
+        await personDao.insertPersons(persons);
+
+        await simulateDoingSomethingElse();
```
vitusortner, line 132: ⚠️ `sleep()` also blocks asynchronous execution. This means that these tests only pass because sqflite spawns another Isolate under the hood? Dart FFI is still sort of unknown to me.
Introducing nondeterministic tests should only be considered as a last resort. Especially when introducing a test that takes 3 seconds. I played with it, trying to find a solution, without any luck. I guess it was the same for you, which made you introduce this function.
I would prefer using `Future.delayed()` instead of `sleep()`, which also seems to work with a delay of only 100 ms.

```suggestion
        await Future<void>.delayed(const Duration(milliseconds: 100));
```

floor/test/test_util/separate_queries.dart
```diff
@@ -0,0 +1,5 @@
+import 'dart:io';
+
+Future<void> simulateDoingSomethingElse([int seconds]) async {
```
vitusortner, line 3: ⚠️Prefer using `Future.delayed()` instead.

floor/test/integration/stream_query_test.dart
```diff
@@ -126,5 +128,36 @@ void main() {
...
+      await personDao.insertPerson(person1);
+      await personDao.insertPerson(person2);
+
+      await database.dogDao.insertDog(dog1);
+
+      await database.dogDao.insertDog(dog2);
+
+      // avoid that delete happens before the re-execution of
+      // the select query for the stream
+      await simulateDoingSomethingElse();
```
vitusortner, line 158: ⚠️Prefer using `Future.delayed()` instead.

vitusortner:
> @vitusortner I noticed that we actually don't test streams quite the right way. Floor's Streams will re-execute the query as soon someone will start to Listen (which is right this way), but that also means that each
> 
> ```dart
> expect(actual, emits([anything]));
> ```
> 
> in the tests just starts listening and reexecutes the query, not actually looking if an event was fired independently. Previously fired events will be discarded if no-one is actually listening. This means that our Stream tests are just checking if the insert/update/delete query got executed, not if the stream fired.
> 
> For the new tests added here I'm using
> 
> ```dart
> expect(actual, emitsInOrder([firstResult,secondResult]));
> ```
> 
> at the start of the test, which will keep the stream open and actually checks the stream output.
> 
> When doing this I noticed that some queries will be executed together before the re-execution of the Stream Query fires, meaning that some events get "lost" and other events will then get a double update.
> 
> For that I introduced a new helper function that will introduce artificial wait times to let floor reexecute the SELECT query before the test starts the next database action. I'll push that soon.

Good find! Thanks

floor/test/integration/view/view_test.dart
```diff
@@ -59,6 +60,24 @@ void main() {
         expect(actual, equals(expected));
       });
 
+// TODO this currently won't work as all `null` results won't be added to the Stream.
+// this is not specific to Views.
```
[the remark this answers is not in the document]
mqus: I'm a bit torn on that one... I think that it can be reasonable to not return null in streams... But I think that it should be documented. I'll open an issue.

floor/test/adapter/query_adapter_test.dart
```diff
@@ -257,5 +259,6 @@ void main() {
         ]),
       );
     });
+    //TODO test view updates every time
```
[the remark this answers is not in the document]
mqus: right, I missed them.

floor/test/integration/view/view_test.dart
```diff
@@ -85,6 +104,48 @@ void main() {
         expect(actual, equals(expected));
       });
 
+      test('query view with all values as stream', () async {
+        final actual = nameDao.findAllNamesAsStream();
+        expect(
+            actual,
+            emitsInOrder(<List<Name>>[
+              [], //initial state
+              [Name('Frank'), Name('Leo')], // after inserting Persons
```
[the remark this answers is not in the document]
mqus: yes because the query has an `ORDER BY`.

floor/test/integration/stream_query_test.dart
```diff
@@ -126,5 +128,36 @@ void main() {
...
+
+      final actual = personDao.findAllDogsOfPersonAsStream(person1.id);
+      expect(
+          actual,
+          emitsInOrder(<List<Dog>>[
+            [], // initial state,
+            [dog1], // after inserting dog1
+            [dog1], // after inserting dog2
+            //[], // after removing person1. Does not work because
+            // ForeignKey-relations are not considered. (TODO)
```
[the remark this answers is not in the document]
mqus: technically this can be solved without adding a query parser because we already know the foreign key relations and their effects (onUpdate/onDelete), but I'll include it in the query parser so I'll link to it.

mqus:
> As Codecov is blocking this PR from getting merged, I suggest that I'll merge it if you give your go?

yes please :)

PR #327, opened 2020-05-11
Migrate OnConflictStrategy to enum

vitusortner:
`OnConflictStrategy` (used by `@update` and `@insert`) now is an enum instead of integers. Thereby, the compiler supports us and the library's users with its type safety.

No migration for end users needed.


vitusortner:
> Btw, I think we should do the same for the ForeignKeyAction.

Yes, I've started working on this and will open another PR soon.

PR #336, opened 2020-05-23
Await database path retrieval

vitusortner:
Without awaiting the path retrieval, the database can't be opened. This issue has been introduced by #326.


vitusortner:
That's what I was wondering as well. I didn't see any warnings when looking through the code. Dart's linter might not pick up un-awaited Futures in conditional statements.

PR #337, opened 2020-05-23
Get rid of unnecessary Flutter binding assertions in tests

vitusortner:


vitusortner:
My intention was to give an overview of which big features are being developed right now and thus, will go into the 1.0 release. We could look into defining a more detailed list in the form of a roadmap. This might also give us a better overview of the current and future state of the library.

Issue #338, opened 2020-05-23
Streams over single values do not emit null values


mqus:
As mentioned in #320, Streams over a single value like the one in the following example do not update if it returns no result:
```dart
@Query('SELECT * FROM Person WHERE id = :id')
Stream<Person> findPersonById(int id);
```
So if the person with the matching id is deleted, the Stream will re-execute the query but will not propagate the null result. 
As this is this way to emulate Rooms behaviour then there are three actions we can take (and I think we should take at least one of them):
1. Document the current behaviour (streams over single events only appear in examples but not in the section about streams
2. Add an option to the `@Query` annotation which enables null propagation/emission
3. Emit null values by default.

Currently I'm leaning towards 1+2 for having backwards compatibility and the option to return empty entries.

vitusortner:
Your first point sounds great. I'll open a PR in a bit to add proper documentation. About the second and third, I would prefer to wait until Dart's non-nullable types arrive as the feature doesn't seem to be too urgent. This way, we can make use of the type system to infer if `null`s should be propagated. Nullable types will have to be marked with `?` whereas non-nullable types will stay the same.

What are your thoughts about this?

```dart
@Query('SELECT * FROM Person WHERE id = :id')
Stream<Person?> findPersonById(int id); // emits null when no person found

@Query('SELECT * FROM Person WHERE id = :id')
Stream<Person> findPersonById(int id); // never emits null
```

PR #343, opened 2020-05-27
Mention missing null propagation in streams

vitusortner:
#338

### DIFF
--- a/README.md
+++ b/README.md
@@ -454,12 +454,11 @@ Future<int> deletePersons(List<Person> person);
 ```
 
 ### Streams
-As already mentioned, queries can not only return a value once when called but also a continuous stream of query results.
-The returned stream keeps you in sync with the changes happening to the database table.
-This feature plays really well with the `StreamBuilder` widget.
+As already mentioned, queries cannot only return values once when called but also continuous streams of query results.
+The returned streams keep you in sync with the changes happening in the database tables.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -454,12 +454,11 @@ Future<int> deletePersons(List<Person> person);
 ```
 
 ### Streams
-As already mentioned, queries can not only return a value once when called but also a continuous stream of query results.
-The returned stream keeps you in sync with the changes happening to the database table.
-This feature plays really well with the `StreamBuilder` widget.
+As already mentioned, queries cannot only return values once when called but also continuous streams of query results.
```
mqus, line 457: This "cannot" seems weird here... I think it was better before.
But I'm not a native, so I don't really know.

vitusortner: My IDE suggested that change. Apparently "cannot" is used more frequently compared to "can not". https://www.grammarly.com/blog/cannot-or-can-not/

mqus: I would usually agree, but the combination with "only" is what triggered me... 
well, it's not that important :smile:

mqus: *evil laugh* found something confirming my position: https://english.stackexchange.com/questions/421944/can-we-use-the-wording-we-can-not-only-but-also

vitusortner: Haha I see 😄

Issue #350, opened 2020-06-01
Bulk inserts lock database and break other stream queries


TheKewlStore:
When using this library to do bulk inserts of upwards of 30000 records, my database gets locked and stream queries that track changes to the daos (particularly ones that aren't even related to the insert).  I attempted to run the insert in a compute or an isolate, but that doesn't work because it doesn't have access to the memory objects it needs.

What did seem to work for me was creating an _insertListVoid function in insertion_adapters.dart which sets noResult to true on the batch commit. Since void functions signal that I don't want a return anyway, I don't see why this wasn't being used already. Then I changed the call to _insertList to _insertListVoid in the void version of insertList. 

I'm submitting a pull request with this change to your library, as it may help others in the future.

TheKewlStore:
https://github.com/vitusortner/floor/pull/351

PR #351, opened 2020-06-01
Adding insertListVoid which passes noResult as true

TheKewlStore:
See https://github.com/vitusortner/floor/issues/350

### DIFF
--- a/floor/lib/src/adapter/insertion_adapter.dart
+++ b/floor/lib/src/adapter/insertion_adapter.dart
@@ -36,7 +36,21 @@ class InsertionAdapter<T> {
     final OnConflictStrategy onConflictStrategy,
   ) async {
     if (items.isEmpty) return;
-    await _insertListVoid(items, onConflictStrategy);
+    final batch = _database.batch();
+    for (final item in items) {
+      batch.insert(
+        _entityName,
+        _valueMapper(item),
+        conflictAlgorithm: onConflictStrategy.asSqfliteConflictAlgorithm(),
+      );
+    }
+
+    await batch.commit(noResult: true);
+
+    if (_changeListener != null) {
+      _changeListener.add(_entityName);
+    }
+  }
   }
 
   Future<int> insertAndReturnId(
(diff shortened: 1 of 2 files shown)


floor/lib/src/adapter/insertion_adapter.dart
```diff
@@ -87,4 +87,24 @@ class InsertionAdapter<T> {
     }
     return result;
   }
+  
+    Future<void> _insertListVoid(
+      final List<T> items,
+      final ConflictAlgorithm conflictAlgorithm,
+      ) async {
```
vitusortner, line 94: 🔍 Formatting seems to be off here
```suggestion
  Future<void> _insertListVoid(
    final List<T> items,
    final ConflictAlgorithm conflictAlgorithm,
  ) async {
```

floor/lib/src/adapter/insertion_adapter.dart
```diff
@@ -36,7 +36,7 @@ class InsertionAdapter<T> {
     final OnConflictStrategy onConflictStrategy,
   ) async {
     if (items.isEmpty) return;
-    await _insertList(items, onConflictStrategy);
+    await _insertListVoid(items, onConflictStrategy);
```
vitusortner, line 39: 💡 You could inline the functionality as it's only used once. You also need to make sure to map `OnConflictStrategy` to sqflite's `ConflictAlgorithm`.

```suggestion
    final batch = _database.batch();
    for (final item in items) {
      batch.insert(
        _entityName,
        _valueMapper(item),
        conflictAlgorithm: onConflictStrategy.asSqfliteConflictAlgorithm(),
      );
    }

    await batch.commit(noResult: true);

    if (_changeListener != null) {
      _changeListener.add(_entityName);
    }
  }
```

TheKewlStore:
Hello, sorry for the delay in getting back to you on this! I accepted the changes you suggested on inlining the function and the formatting issues. I also adjusted the test you mentioned (by ensuring that it checks for the noResult parameter to be true). Let me know if there are any other changes necessary!

vitusortner:
There was a bracket too much, which I got rid of. Thanks 🙏

PR #353, opened 2020-06-02
Fix String escaping and add test

### LINKED ISSUE #352: Can't compare fields against blank
I'm not sure if its a Floor's issue, anyway I'm reporting here just to check.

Anytime I try to compare a field against blank (''), the generated comparison breaks the SQL. For instance:
```sql
select * from Customer c where c.city = '' and c.state = ?
```
The output is: 

```sql
select * from customer c where c.city = and c.state = 'PR'
```

mqus:
Fixes #352 by escaping Strings the right way. Technically we should also remove the altering of the query we do in the `query_method_processor` (replacing newlines and multiple sequential spaces) but I think those changes are better placed in the queryparser (which is in the works but not quite ready for a PR :wink:  )

### DIFF
--- a/floor_generator/lib/writer/query_method_writer.dart
+++ b/floor_generator/lib/writer/query_method_writer.dart
@@ -126,7 +126,10 @@ class QueryMethodWriter implements Writer {
     @nullable final String arguments,
     @nonNull final String mapper,
   ) {
-    final parameters = StringBuffer()..write("'${_queryMethod.query}', ");
+    final parameters = StringBuffer()
+      ..write(literalString(_queryMethod.query))
+      ..write(', ');
+
     if (arguments != null) parameters.write('arguments: $arguments, ');
     parameters.write('mapper: $mapper');
 
--- a/floor_generator/lib/writer/query_method_writer.dart
+++ b/floor_generator/lib/writer/query_method_writer.dart
@@ -116,7 +116,7 @@ class QueryMethodWriter implements Writer {
 
   @nonNull
   String _generateNoReturnQuery(@nullable final String arguments) {
-    final parameters = StringBuffer()..write("'${_queryMethod.query}'");
+    final parameters = StringBuffer()..write('r""" ${_queryMethod.query} """');
     if (arguments != null) parameters.write(', arguments: $arguments');
     return 'await _queryAdapter.queryNoReturn($parameters);';
   }
@@ -127,7 +127,7 @@ class QueryMethodWriter implements Writer {
     @nonNull final String mapper,
   ) {
     final parameters = StringBuffer()
-      ..write(literalString(_queryMethod.query))
+      ..write('r""" ${_queryMethod.query} """')
       ..write(', ');
 
     if (arguments != null) parameters.write('arguments: $arguments, ');
@@ -147,7 +147,8 @@ class QueryMethodWriter implements Writer {
   ) {
     final queryableName = _queryMethod.queryable.name;
     final isView = _queryMethod.queryable is View;
-    final parameters = StringBuffer()..write("'${_queryMethod.query}', ");
+    final parameters = StringBuffer()
+      ..write('r""" ${_queryMethod.query} """, ');
     if (arguments != null) parameters.write('arguments: $arguments, ');
     parameters
       ..write("queryableName: '$queryableName', ")
--- a/floor/test/extension/on_conflict_strategy_extensions_test.dart
+++ b/floor/test/extension/on_conflict_strategy_extensions_test.dart
@@ -46,7 +46,7 @@ void main() {
     });
 
     test('falls back to abort when null', () {
-      final actual = null.asSqfliteConflictAlgorithm();
+      final ConflictAlgorithm actual = null.asSqfliteConflictAlgorithm();
 
       expect(actual, equals(ConflictAlgorithm.abort));
     });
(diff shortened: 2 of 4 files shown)


vitusortner:
Thanks for the fix and the quick update on the query parser!
I totally agree that we should take care about the transformed queries. They've been causing too much trouble.

floor_generator/lib/writer/query_method_writer.dart
```diff
@@ -126,7 +126,10 @@ class QueryMethodWriter implements Writer {
     @nullable final String arguments,
     @nonNull final String mapper,
   ) {
-    final parameters = StringBuffer()..write("'${_queryMethod.query}', ");
+    final parameters = StringBuffer()
+      ..write(literalString(_queryMethod.query))
```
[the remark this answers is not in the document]
mqus: > did you check out Dart's raw strings marked by a leading r? e.g. `r'string'`

No, but I definitely should, thanks! Edit: I did. 

> what if we would get rid of the multi to single line parsing which would allow people to use double quotes and thereby, would solve the initial problem described in #352?

What do you mean?

mqus:
I also changed the other places where the query is inserted into the queryAdapter

floor_generator/lib/writer/query_method_writer.dart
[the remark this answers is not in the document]
vitusortner: We currently transform any query supplied to `@Query()` into a single-line string using single quotes. Even when the initial query was defined with double quotes, we get rid of them.

Now to the problem mentioned in #352. The user defined a multiline query with three leading and trailing single quotes. Using single quotes in such a string to define an empty string isn't a problem at all. The same applies to strings surrounded by double quotes.

As you found a solution for the problem, I suggest to keep the multi to single-line transformation issues in mind and get this PR merged without these changes.

floor_generator/lib/writer/query_method_writer.dart
```diff
@@ -116,7 +116,7 @@ class QueryMethodWriter implements Writer {
 
   @nonNull
   String _generateNoReturnQuery(@nullable final String arguments) {
-    final parameters = StringBuffer()..write("'${_queryMethod.query}'");
+    final parameters = StringBuffer()..write('r""" ${_queryMethod.query} """');
```
vitusortner, line 119: ❓ Did you add the leading and trailing space on purpose?
```suggestion
    final parameters = StringBuffer()..write('r"""${_queryMethod.query}"""');
```

floor_generator/lib/writer/query_method_writer.dart
```diff
@@ -126,7 +126,10 @@ class QueryMethodWriter implements Writer {
     @nullable final String arguments,
     @nonNull final String mapper,
   ) {
-    final parameters = StringBuffer()..write("'${_queryMethod.query}', ");
+    final parameters = StringBuffer()
+      ..write('r""" ${_queryMethod.query} """')
```
vitusortner, line 130: ❓ Did you add the leading and trailing space on purpose?
```suggestion
      ..write('r"""${_queryMethod.query}"""')
```

floor_generator/lib/writer/query_method_writer.dart
```diff
@@ -144,7 +147,8 @@ class QueryMethodWriter implements Writer {
   ) {
     final queryableName = _queryMethod.queryable.name;
     final isView = _queryMethod.queryable is View;
-    final parameters = StringBuffer()..write("'${_queryMethod.query}', ");
+    final parameters = StringBuffer()
+      ..write('r""" ${_queryMethod.query} """, ');
```
vitusortner, line 151: ❓ Did you add the leading and trailing space on purpose?
```suggestion
      ..write('r"""${_queryMethod.query}""", ')
```

floor_generator/lib/writer/query_method_writer.dart
mqus: Yes, if the query contains a `""` at the end then it will be concatenated to `"""""`, which dart then will interpret as `""" ""`, which is wrong. so I put a space at the front (shouldn't matter, just in case) and a space at the end. Both spaces should be permitted by sqlite's syntax.

Example query (although `"`s there are not really valid sql, but we shouldn't change them anyways.):
```sql
SELECT * FROM table WHERE x=""
```
(`"` quotes in sqlite are used for identifiers https://www.sqlite.org/lang_keywords.html)
I added a test case for the trailing `'` quotes before but forgot to add one for the other ones. I'll be adding another one for the `"` quotes now. I think I'll also add one for backtick (` ` ` ) quotes, even if they're not causing trouble right now.

floor_generator/lib/writer/query_method_writer.dart
mqus: See above

floor_generator/lib/writer/query_method_writer.dart
mqus: See above

mqus:
I just ran the integration tests and it seems that we can't use `r""" """` quotes because we reference dart values when using IN clauses. But these are just interpreted literally in raw strings.
There are two solutions:
- not using raw strings and just escaping it (like the first iteration did)
- using raw strings but inserting the variables for the in clause in a special way.

Anyways, `IN` clauses are also something which I want to adress in the queryparser, mainly because we are currently able to be sql-injected via them (although the risk and the consequences are small).

mqus:
Because of the issues mentioned above, I will close this PR in favor of #361.

PR #359, opened 2020-06-14
Change ForeignKeyAction to enum in the generator

mqus:
This changes ForeignKeyAction to be an enum instead of a string. This allows us to be more strict when processing the annotation as well as be more confident when dealing with annotations.

This is a preparation for the sqlanalyzer, to avoid matching strings when converting floor entities to moor/sqlparser tables.


floor_annotation/lib/src/foreign_key.dart
```diff
@@ -41,7 +41,7 @@ abstract class ForeignKeyAction {
   /// When a parent key is modified or deleted from the database, no special
   /// action is taken. This means that SQLite will not make any effort to fix
   /// the constraint failure, instead, reject the change.
-  static const noAction = 1;
+  static const noAction = 0;
```
vitusortner, line 44: 💡 We could actually use an enum through all layers. In order to retrieve the enum values from `ForeignKey.onUpdate` and `ForeignKey.onDelete` annotation fields, the following function would do the trick. This would save us the conversion from `int` to `ForeignKeyAction` which would reduce complexity and future maintenance costs. What are your thoughts about this approach?

```dart
extension on DartObject {
  annotations.ForeignKeyAction toForeignKeyAction() {
    final enumValueString = toEnumValueString(); // defined in DartObjectExtension
    return annotations.ForeignKeyAction.values.singleWhere(
        (foreignKeyAction) => foreignKeyAction.toString() == enumValueString);
  }
}

// EntityProcessor._getForeignKeys
final onUpdateForeignKeyAction = foreignKeyObject
    .getField(ForeignKeyField.onUpdate)
    ?.toForeignKeyAction();
```

floor_generator/lib/misc/foreign_key_action.dart
```diff
@@ -18,9 +16,21 @@ abstract class ForeignKeyAction {
         return 'SET DEFAULT';
       case ForeignKeyAction.cascade:
         return 'CASCADE';
-      case ForeignKeyAction.noAction:
       default:
-        return 'NO ACTION';
+        assert(false, 'Unexpected enum case $this');
```
vitusortner, line 20: `assert`s only are enabled in debug mode. Was this intentional? https://dart.dev/guides/language/language-tour#assert

mqus: I wanted to throw an Error without overhead (if anything ever throws this error, it is our fault and not the one of library users)... assert seems to be the wrong way then... Do you have an idea?

floor_annotation/lib/src/foreign_key.dart
mqus: Is fine for me!

mqus:
I put the extension on `annotations.ForeignKeyAction` in the extensions folder (mainly because there is nothing else left in the file) and let `value_object/foreign_key.dart` re-export `annotations.ForeignKeyAction` for easy use in the writer.

floor_generator/lib/misc/foreign_key_action.dart
vitusortner: As far as I know, `assert` throws an `AssertError`. So in the end, it's just an `Error`.
I've started using the built-in errors such as `UnsupportedError`, `StateError` or `ArgumentError` recently, which I like.

floor_generator/lib/misc/extension/foreign_key_action.dart
```diff
@@ -18,9 +15,8 @@ abstract class ForeignKeyAction {
         return 'SET DEFAULT';
       case ForeignKeyAction.cascade:
         return 'CASCADE';
-      case ForeignKeyAction.noAction:
       default:
-        return 'NO ACTION';
+        return null;
```
vitusortner, line 19: 💡 We should change the `@nonNull` annotation to `@nullable` if returning `null` is really what we want here. Have you thought of throwing an `Error` instead? I'm thinking about this approach because we're in full control of the cases (`ForeignKeyAction` enum) so we *should* never see the error in "production".

Some background about the `@nonNull` and `@nullable` annotations: At some point back in time, I heard that the Dart team at Google is working on making the language null-safe. The time is about to arrive as [non-nullable types are currently in preview](https://medium.com/dartlang/announcing-sound-null-safety-defd2216a6f3) and will be released later this year. In order for us to have an easy migration, the annotations will help 🙂

floor_generator/lib/processor/entity_processor.dart
```diff
@@ -205,4 +204,24 @@ class EntityProcessor extends QueryableProcessor<Entity> {
...
+      return annotations.ForeignKeyAction.noAction;
+    }
+
+    final fka = field.toForeignKeyAction();
+    if (fka == null) {
+      // field was not defined correctly
+      throw _processorError.wrongForeignKeyAction(field, isUpdate);
+    }
+    return fka;
+  }
```
vitusortner, line 226: 💡 We could improve this abstraction by finding a way to not need to pass a boolean to it. We could, for instance, just pass the field name to it instead.

floor_generator/lib/misc/extension/foreign_key_action.dart
mqus: This code is now throwing an ArgumentError.

floor_generator/lib/misc/extension/foreign_key_action.dart
```diff
@@ -1,15 +1,12 @@
...
-  static const restrict = 2;
-  static const setNull = 3;
-  static const setDefault = 4;
-  static const cascade = 5;
-
+extension ToSQL on ForeignKeyAction {
   @nonNull
-  static String getString(final int action) {
-    switch (action) {
+  String get toSQL {
```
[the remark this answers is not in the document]
mqus: here you go :)

floor_generator/lib/misc/extension/dart_object_extension.dart
```diff
@@ -1,14 +1,26 @@
...
 
 extension DartObjectExtension on DartObject {
   String toEnumValueString() {
     final interfaceType = type as InterfaceType;
     final enumValue = interfaceType.element.fields
         .where((element) => element.isEnumConstant)
         .map((fieldElement) => fieldElement.name)
-        .singleWhere((valueName) => getField(valueName) != null);
+        .singleWhere((valueName) => getField(valueName) != null,
+            orElse: () => null);
```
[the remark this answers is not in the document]
mqus: The context for a good error message is not available here, so I'm adding a callback for `orElse`.

floor_generator/lib/misc/extension/dart_object_extension.dart
```diff
@@ -1,14 +1,26 @@
...
         .map((fieldElement) => fieldElement.name)
-        .singleWhere((valueName) => getField(valueName) != null);
+        .singleWhere((valueName) => getField(valueName) != null,
+            orElse: orElse);
 
     return '$interfaceType.$enumValue';
   }
+
+  ForeignKeyAction toForeignKeyAction({@required ForeignKeyAction orElse()}) {
+    final enumValueString = toEnumValueString(orElse: () => null);
```
vitusortner, line 21: 💡 As when the `orElse` value is used here, we could return early and just use the value produced from the `orElse` function parameter.

```suggestion
    final enumValueString = toEnumValueString(orElse: () => null);
    if (enumValueString == null) return orElse();
```

mqus: I don't think `enumValueString` can ever be `null`, because `toEnumValueString` will always return `'$interfaceType.$enumValue'`, or in the orElse case: `'$interfaceType.null'`. So this short-circuit won't work.

mqus: But I see that this `orElse` parameter in `toEnumValueString` is misleading. I should change that.

vitusortner:
@mqus I'll look into getting this one merged if you don't mind?

vitusortner (approved):

mqus:
Thank you!

Issue #358, opened 2020-06-14
WITHOUT ROWID table support


John-He-928:
It would be great to support creating [WITHOUT ROWID tables](https://www.sqlite.org/withoutrowid.html) using the `@Entity` annotation like `@Entity(withoutRowid: true)`.

WITHOUT ROWID table annotation should be easy to implement by just appending `WITHOUT ROWID` in the CREATE TABLE statement. It may be useful in creating small tables with primary keys other than `INTEGER PRIMARY KEY`. One example would be Many-to-Many relationship table:

```sql
CREATE TABLE Gallery (
    id INTEGER PRIMARY KEY, 
    url TEXT
    /* some other fields */ );

CREATE TABLE Tag (
    id INTEGER PRIMARY KEY, 
    name TEXT
    /* some other fields */ );

CREATE TABLE GalleryTag (
    galleryId INTEGER REFERENCES Gallery,
    tagId INTEGER REFERENCES Tag,
    PRIMARY KEY(galleryId, tagId)
) WITHOUT ROWID;
```

mqus:
I think this is a good feature to have and shouldn't take much effort to implement. `WITHOUT ROWID` does only work in sqlite but I think we don't have any plans to support other SQL dialects.

IMHO we can definitely accept pull requests for this.

vitusortner:
Closed by #365

PR #365, opened 2020-06-24
Add support for WITHOUT ROWID tables

John-He-928:
See #358 

Added `withoutRowid` parameter to `@Entity` annotation, which appends `WITHOUT ROWID` to the CREATE TABLE statement if set to true.

Simple test cases included.

### DIFF
--- a/floor_generator/lib/value_object/entity.dart
+++ b/floor_generator/lib/value_object/entity.dart
@@ -110,6 +110,6 @@ class Entity extends Queryable {
 
   @override
   String toString() {
-    return 'Entity{classElement: $classElement, name: $name, fields: $fields, primaryKey: $primaryKey, foreignKeys: $foreignKeys, indices: $indices, constructor: $constructor}';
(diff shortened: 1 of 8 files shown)


floor_generator/lib/value_object/entity.dart
```diff
@@ -100,7 +105,8 @@ class Entity extends Queryable {
       primaryKey.hashCode ^
       foreignKeys.hashCode ^
       indices.hashCode ^
-      constructor.hashCode;
+      constructor.hashCode ^
+      withoutRowid.hashCode;
 
   @override
   String toString() {
```
mqus, line 112: Please also change the `toString` method.

mqus:
So first of all, thank you for your contribution and sorry for not looking at it sooner. I have mainly two minor issues with your code: The first is the `toString` method (see the comment above) and the second is the following:
I'm wondering if it would make sense to add a check for points 1 and 3 in the [sqlite docs](https://sqlite.org/withoutrowid.html#differences_from_ordinary_rowid_tables) , meaning that `WITHOUT ROWID` tables should have a primary key that is not auto-incremented.

Apart from that, I'm pretty happy with your changes. But @vitusortner  should probably take a look.

John-He-928:
@mqus Thank you for the review. I will update the code :)

vitusortner:
Thanks for your contribution and thanks for your patience!

PR #381, opened 2020-07-25
Check transaction method return types and allow non-void returns

### LINKED ISSUE #379: Generated method without return statement
I have the following method in a class called `SimulationDao` to save a simulation and its items: 

```
@transaction
  Future<int> saveWithItems(SimulationEntity simulation) async {
    final savedId = await this.save(simulation);
    simulation.id = savedId;
    simulation.items.forEach((element) => element.simulationId = savedId);
    DaoFactory.instance.simulationItemDao.saveAll(simulation.items);
    return savedId;
  }
```

But the generated method has no return statement: 
```
  @override
  Future<int> saveWithItems(SimulationEntity simulation) async {
    if (database is sqflite.Transaction) {
      await super.saveWithItems(simulation); 
    } else {
      await (database as sqflite.Database)
          .transaction<void>((transaction) async {
        final transactionDatabase = _$AppDatabase(changeListener)
          ..database = transaction;
        await transactionDatabase.simulationDao.saveWithItems(simulation);
      });
    }
  }
```

mqus:
Closes #379

### DIFF
--- a/README.md
+++ b/README.md
@@ -529,7 +529,7 @@ StreamBuilder<List<Person>>(
 
 ### Transactions
 Whenever you want to perform some operations in a transaction you have to add the `@transaction` annotation to the method.
-It's also required to add the `async` modifier. These methods can only return `Future<>`.
+It's also required to add the `async` modifier. These methods have to return a `Future`.
 
 ```dart
 @transaction
--- a/README.md
+++ b/README.md
@@ -529,7 +529,7 @@ StreamBuilder<List<Person>>(
 
 ### Transactions
 Whenever you want to perform some operations in a transaction you have to add the `@transaction` annotation to the method.
-It's also required to add the `async` modifier. These methods can only return `Future<void>`.
+It's also required to add the `async` modifier. These methods can only return `Future<>`.
 
 ```dart
 @transaction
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -529,7 +529,7 @@ StreamBuilder<List<Person>>(
 
 ### Transactions
 Whenever you want to perform some operations in a transaction you have to add the `@transaction` annotation to the method.
-It's also required to add the `async` modifier. These methods can only return `Future<void>`.
+It's also required to add the `async` modifier. These methods can only return `Future<>`.
```
vitusortner, line 532: 🔍 We could rephrase the second sentence in order to make the necessity of returning `Future`s sound less like a limitation but more a constraint or feature.

```suggestion
It's also required to add the `async` modifier. These methods have to return a `Future`.
```

floor/README.md
```diff
@@ -529,7 +529,7 @@ StreamBuilder<List<Person>>(
 
 ### Transactions
 Whenever you want to perform some operations in a transaction you have to add the `@transaction` annotation to the method.
-It's also required to add the `async` modifier. These methods can only return `Future<void>`.
+It's also required to add the `async` modifier. These methods can only return `Future<>`.
```
vitusortner, line 532: https://github.com/vitusortner/floor/pull/381#discussion_r463192085

mqus:
```suggestion
It's also required to add the `async` modifier. These methods can only return a `Future`.
```

Issue #397, opened 2020-09-12
Which version is planned to support fts


jiechic:
We use a room that supports fts on the native development of android. We hope to be able to migrate to flutter. Look for a flutter package similar to room on flutter. We found you
We would like to ask if there are plans to support fts

mqus:
Could you please expand on what fts is? It doesn't ring a bell immediately for me, sorry :)

And following that, since we are not room, you could also describe what you feel is missing from rooms fts feature(if that is possible) or what could be better. Maybe we could expand on that.

mqus:
So I actually looked it up now (should have done that before commenting, sry) and fts(full text search) is an optional sqlite module. 

Afaict, this situation is the same as with the json extension: sqflite(our sqlite database driver) does not support it, so we also can't support it right now. I'm also not sure where the issue is(if this is already supported for all platforms that sqflite supports or if e.g. iOS does not include the module)

jiechic:
In the [official sqlite documentation](https://www.sqlite.org/fts3.html),
`FTS4 is an enhancement to FTS3. FTS3 has been available since SQLite version 3.5.0 (2007-09-04) The enhancements for FTS4 were added with SQLite version 3.7.4 (2010-12-07).`

on Android:
I found the [official document of android](https://developer.android.com/reference/android/database/sqlite/package-summary.html)
since android 11 was beging sqlite version 3.7，I see [test record](https://stackoverflow.com/questions/2421189/version-of-sqlite-used-in-android/4377116#4377116) since android 11 has beging sqlite version 3.7.4

on iOS
i dont't found Any information on official document
i found [test record](https://github.com/yapstudios/YapDatabase/wiki/SQLite-version-(bundled-with-OS)) for get some information
since iOS 5.1.1 is beging sqlite version 3.7.7

i found sqflite minversion is android 16 and iOS 8
so，sqflite is support fts3 and fts4，but sqflite don't support fts5 on all android or ios version ，you can see issue in [sqflite](https://github.com/tekartik/sqflite/issues/105)

Can you consider adding support for fts3 and fts4 like room？
[room fts3](https://developer.android.com/reference/androidx/room/Fts3)
[room fts4](https://developer.android.com/reference/androidx/room/Fts4)

jiechic:
and i see dart office package [sqlite3](https://pub.dev/packages/sqlite3) has new version for sqlite，i can support fts3、 fts4、fts5 on any platforms but not only android ，ios
[...]

PR #407, opened 2020-10-15
Add Full-text Search support

jiechic:
Add Fts3 & Fts4 support


jiechic:
what wrong with codecov/patch ?
i don't know how to fix?

jiechic:
resolve issue #397

vitusortner:
Thanks for your contribution 🙏 I'll look into reviewing it ASAP.

floor_generator/lib/value_object/entity.dart
```diff
@@ -45,7 +48,13 @@ class Entity extends Queryable {
 
     final withoutRowidClause = withoutRowid ? ' WITHOUT ROWID' : '';
 
-    return 'CREATE TABLE IF NOT EXISTS `$name` (${databaseDefinition.join(', ')})$withoutRowidClause';
+    if (fts == null) {
+      return 'CREATE TABLE IF NOT EXISTS `$name` (${databaseDefinition.join(', ')})$withoutRowidClause';
+    } else {
+      final ftsType = fts.type ?? '';
+      final ftsTokenizer = fts.tokenizer ?? '';
+      return 'CREATE VIRTUAL TABLE IF NOT EXISTS `$name` USING $ftsType(${databaseDefinition.join(', ')}, tokenize=$ftsTokenizer)$withoutRowidClause';
```
vitusortner, line 56: ❓ Does FTS work without a `rowid`?

```suggestion
      return 'CREATE VIRTUAL TABLE IF NOT EXISTS `$name` USING $ftsType(${databaseDefinition.join(', ')}, tokenize=$ftsTokenizer)';
```

floor_annotation/lib/src/fts.dart
```diff
@@ -0,0 +1,34 @@
+class Fts {
```
vitusortner, line 1: 💡 FTS4 has some additional features that FTS3 doesn't support. Such as setting an external content table to be used by the FTS table. It might make sense to create two separate types: `FTS3` and `FTS4` as it's done in Room (see links below). We don't have to support all of these additional features initially but having two classes might allow us to be more flexible in the future.
Having two FTS classes also requires some changes to `Entity` because it would either need to understand both the `FTS3` and `FTS4`types or we list FTS out of the `Entity` annotation as done in Room. What are your thoughts about this?

https://developer.android.com/reference/androidx/room/Fts3
https://developer.android.com/reference/androidx/room/Fts4

floor_annotation/lib/src/fts.dart
```diff
@@ -0,0 +1,34 @@
...
+
+  ///The name of the default tokenizer used on FTS tables
+  static const simple = 'simple';
+
+  ///The name of the tokenizer based on the Porter Stemming Algorithm
+  static const porter = 'porter';
+
+  ///The name of a tokenizer implemented by the ICU library.
+  static const icu = 'icu';
+}
```
vitusortner, line 34: 💡 We could use enums here instead of strings to benefit from type safety. We've done this with `OnConflictStrategy` as you can see here https://github.com/vitusortner/floor/blob/c5925af0f8ab319cd6e8436143bc4c7aace9b2d1/floor_annotation/lib/src/on_conflict_strategy.dart and here https://github.com/vitusortner/floor/blob/c5925af0f8ab319cd6e8436143bc4c7aace9b2d1/floor_generator/lib/processor/update_method_processor.dart#L72.

example/lib/database.dart
```diff
@@ -1,13 +1,17 @@
...
+import 'package:example/mail_dao.dart';
 import 'package:example/task.dart';
 import 'package:example/task_dao.dart';
 import 'package:floor/floor.dart';
 import 'package:sqflite/sqflite.dart' as sqflite;
 
 part 'database.g.dart';
 
-@Database(version: 1, entities: [Task])
+@Database(version: 1, entities: [Task, Mail])
```
vitusortner, line 12: 💡 I would suggest keeping the example app as simple as possible. Thereby, we could move the `Mail` entity and `MailDao` into the integration test directory of floor. Additionally, it would be nice to have some actual integration tests that assert the functionality of FTS. You can find some examples here https://github.com/vitusortner/floor/tree/develop/floor/test/integration.

floor_annotation/pubspec.lock
```diff
@@ -5,7 +5,7 @@ packages:
     dependency: "direct main"
     description:
       name: meta
-      url: "https://pub.dartlang.org"
+      url: "https://pub.flutter-io.cn"
```
vitusortner, line 8: ❓ Is there a way for us to keep referring to `pub.dartlang.org` for dependencies even when developing from China?

vitusortner (changes requested):

floor_annotation/pubspec.lock
jiechic: sorry to change pub.dartlang.org，in my repository，the pubspec.lock is not on version control system，i forget to check this file

floor_generator/lib/value_object/entity.dart
jiechic: ths，next commit will change

floor_annotation/lib/src/fts.dart
jiechic: at last time ，i didn't known how to get a new annotation on classElement
i change to code like room now

floor_annotation/lib/src/fts.dart
jiechic: yes,i will fix

example/lib/database.dart
jiechic: i delete mail and maildao on new commit
next time i will write same actual integration tests that assert the functionality of FTS

jiechic:
why [ci/floor-generator](https://github.com/vitusortner/floor/pull/407/checks?check_run_id=1441039681) fail false but my local is success
image

jiechic:
i change commit for fix for support some you sugguest，not is ok？

jiechic:
any update for review？

vitusortner (commented):
I was only able to partly review the PR for now. I'll review the rest ASAP.

floor_generator/lib/misc/fts_action.dart
```diff
@@ -0,0 +1,26 @@
+import 'package:floor_generator/misc/annotations.dart';
```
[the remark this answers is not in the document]
jiechic: i forgot to delete this file，i fix in new commit

floor_generator/test/value_object/fts_test.dart
```diff
@@ -0,0 +1,57 @@
...
+      expect(usingOptionActual, equals(usingOptionExpected));
+
+      final tableCreateActual = fts.tableCreateOption();
+
+      const tableCreateExpected = 'tokenize=simple ';
+
+      expect(tableCreateActual, equals(tableCreateExpected));
+    });
+
+    test('Fts3 Definition', () {
```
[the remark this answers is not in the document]
jiechic: already fix

jiechic:
alread check and fix，please review again

vitusortner (approved):
Sorry for the late reply and thanks for your contribution. This is looking really good. I'll extend the project's documentation so that your contribution is easily discoverable.

Issue #375, opened 2020-07-18
Null safety


noordawod:
Hi again.

Is there a plan to support null safety for the generator sometime in the future?

Thanks.

mqus:
I assume that you're referring to the [null safety of dart](https://medium.com/dartlang/announcing-sound-null-safety-defd2216a6f3) and how we will interact with non-nullable types in the code analysis and generation. And (speaking for me) there absolutely is one. 

The sqlparser integration(still WIP but complete feature-wise) already supports type derivation of (non-)null query parameters and type checking of (non-)nullable returns. Implementing this should mostly consist of more tests and maybe ten lines of code in the actual generator. Columns of entities and views already have a nullability assigned to it through the `@ColumnInfo` annotation. I'm not sure on how to proceed there yet, but I think we should use some combination of type/deprecation warnings and type checking errors while slowly phasing out the `nullable` parameter. This also depends on how optional the non-nullable types are (this is something I haven't checked out yet).

Let's track this feature here.

I've written a lot of Java and a bit of Rust before I started dabbling in Dart and the decision to have nullable types (even primitives like booleans!) seemed really backwards to me, so I'm very happy they found a solution and I am all for also supporting it in floor.


Btw, internally, vitus already introduced [`@nullable` and `@nonNull`](https://github.com/vitusortner/floor/blob/develop/floor_generator/lib/misc/annotations.dart) for annotating (non-)null returns of internal functions, just to get us ready for also using nonnullable types inside the floor codebase. This won't matter for users of floor but will probably ease our pain when migrating a lot more, even if the annotations were only cosmetically until now.

vitusortner:
I do agree with everything @mqus mentioned. There is high interest in supporting Dart's non-nullable types!

mqus:
Bookmark with lots of useful details: https://dart.dev/null-safety/understanding-null-safety

vitusortner:
Work on null safety support has been started here https://github.com/vitusortner/floor/tree/nullsafety

dkunzler:
Nice that you are working on it. 
Would you mind publishing your branch as an unstable version to pub because I want to try it out.
If I include the dependency like so:
```
  floor:
    git:
      url: https://github.com/vitusortner/floor
      path: floor
      ref: nullsafety
```
[...]

PR #435, opened 2020-11-21
Make floor runtime dependency null safe

vitusortner:
Part of #375.


.github/workflows/ci.yml
```diff
@@ -57,6 +57,7 @@ jobs:
       - name: Run tests
         if:
         working-directory: floor_generator
+        # remove --no-sound-null-safety once all dependencies have migrated
```
vitusortner, line 60: Add TODO

floor_generator/lib/misc/extension/dart_type_extension.dart
```diff
@@ -0,0 +1,18 @@
...
+import 'package:analyzer/dart/element/nullability_suffix.dart';
+import 'package:analyzer/dart/element/type.dart';
+
+extension DartTypeExtension on DartType {
+  // TODO #375 test
+  bool get isNullable {
+    final nullabilitySuffix = this.nullabilitySuffix;
+    switch (nullabilitySuffix) {
+      case NullabilitySuffix.question:
+      case NullabilitySuffix.star: // support legacy code without non-nullables
```
vitusortner, line 12: This will cause problems the way `isNullable` is used right now because it won't allow usages from legacy code as we're always throwing.

Issue #440, opened 2020-12-07
DatabaseView throws error when starting with WITH statement


themadmrj:
`@DatabaseView("WITH lastMessages as (SELECT *, MAX(creationTimestamp) creationTimestamp FROM chat_message GROUP BY chatid) SELECT chat.*, lastMessages.* FROM chat INNER JOIN lastMessages ON chat.id = lastMessages.chatid")`
`class ChatWithLastMessage { ... }`

This query throws an error: `There is no SELECT query defined on the database view ChatWithLastMessage.`

Rewriting to `@DatabaseView("SELECT chat.*, (SELECT *, MAX(creationTimestamp) creationTimestamp FROM chat_message GROUP BY chatid) as lastMessages FROM chat INNER JOIN lastMessages ON chat.id = lastMessages.chatid")` seems to work.

vitusortner:
Thanks for your report. Common table expressions (temporary views created with "WITH") are currently not supported. I might be able to look into this in the next days. If you're interested in contributing, you can take a look at the `ViewProcessor` which currently checks if a view query starts with `SELECT`. You can find the relevant code here: https://github.com/vitusortner/floor/blob/1a49c0f3c3cc49e32846e15472ce13ecea2d6ce2/floor_generator/lib/processor/view_processor.dart#L49

A solution could be to allow queries staring with `WITH` that additionally contain at least one `SELECT`.

themadmrj:
So I can basically just make a pull request changing:
```
if (query == null || !query.trimLeft().toLowerCase().startsWith('select')) {
  throw _processorError.missingQuery;
}
```
to
```
if (query == null || (!query.trimLeft().toLowerCase().startsWith('select') && !(query.trimLeft().toLowerCase().startsWith('with') && query.toLowerCase().contains('select')))) {
  throw _processorError.missingQuery;
}

?
```

vitusortner:
Sounds perfect. Please don't forget to add tests 👌

themadmrj:
> Sounds perfect. Please don't forget to add tests 👌

I'll give it a shot... I've never actually written tests :p

vitusortner:
Closed by #443

PR #443, opened 2020-12-09
Added support for WITH statements for DatabaseViews

themadmrj:
Solves #440.

### DIFF
--- a/floor_generator/lib/processor/view_processor.dart
+++ b/floor_generator/lib/processor/view_processor.dart
@@ -64,8 +64,3 @@ extension on String {
         'select'.allMatches(lowerCasedString).length >= 2;
   }
 }
-      throw _processorError.missingQuery;
-    }
-    return query;
-  }
-}
--- a/floor_generator/lib/processor/view_processor.dart
+++ b/floor_generator/lib/processor/view_processor.dart
@@ -46,11 +46,24 @@ class ViewProcessor extends QueryableProcessor<View> {
         .getField(AnnotationField.viewQuery)
         ?.toStringValue();
 
-    final lowerQuery = query.toLowerCase();
-    if (query == null ||
-        (!lowerQuery.trimLeft().startsWith('select') &&
-            !(lowerQuery.trimLeft().startsWith('with') &&
-                lowerQuery.contains('select')))) {
+    if (query == null || !(query.isSelectQuery || query.isCteWithSelect)) {
+      throw _processorError.missingQuery;
+    }
+    return query;
+  }
+}
+
+extension on String {
+  bool get isSelectQuery => toLowerCase().trimLeft().startsWith('select');
+
+  /// whether the string is a common table expression
+  /// followed by a `SELECT` query
+  bool get isCteWithSelect {
+    final lowerCasedString = toLowerCase();
+    return lowerCasedString.trimLeft().startsWith('with') &&
+        'select'.allMatches(lowerCasedString).length >= 2;
+  }
+}
       throw _processorError.missingQuery;
     }
     return query;
--- a/floor_generator/lib/processor/view_processor.dart
+++ b/floor_generator/lib/processor/view_processor.dart
@@ -46,7 +46,11 @@ class ViewProcessor extends QueryableProcessor<View> {
         .getField(AnnotationField.viewQuery)
         ?.toStringValue();
 
-    if (query == null || !query.trimLeft().toLowerCase().startsWith('select')) {
+    final lowerQuery = query.toLowerCase();
+    if (query == null ||
+        (!lowerQuery.trimLeft().startsWith('select') &&
+            !(lowerQuery.trimLeft().startsWith('with') &&
+                lowerQuery.contains('select')))) {
       throw _processorError.missingQuery;
     }
     return query;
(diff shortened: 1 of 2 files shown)


floor_generator/lib/processor/view_processor.dart
```diff
@@ -46,7 +46,11 @@ class ViewProcessor extends QueryableProcessor<View> {
         .getField(AnnotationField.viewQuery)
         ?.toStringValue();
 
-    if (query == null || !query.trimLeft().toLowerCase().startsWith('select')) {
+    final lowerQuery = query.toLowerCase();
+    if (query == null ||
+        (!lowerQuery.trimLeft().startsWith('select') &&
+            !(lowerQuery.trimLeft().startsWith('with') &&
+                lowerQuery.contains('select')))) {
```
vitusortner, line 53: 💡 We could think about extracting these checks into functions similar to shown in the following snippet. I realized that a `WITH` query needs **at least two** `SELECT`s. One in the `WITH` statement and another one in the statement that provides the database view.

```suggestion
    if (query == null || !(query.isSelectQuery || query.isCteWithSelect)) {
      throw _processorError.missingQuery;
    }
    return query;
  }
}

extension on String {
  bool get isSelectQuery => toLowerCase().trimLeft().startsWith('select');

  /// whether the string is a common table expression
  /// followed by a `SELECT` query
  bool get isCteWithSelect {
    final lowerCasedString = toLowerCase();
    return lowerCasedString.trimLeft().startsWith('with') &&
        'select'.allMatches(lowerCasedString).length >= 2;
  }
}
```

floor_generator/test/processor/view_processor_test.dart
```diff
@@ -36,6 +36,36 @@ void main() {
...
+    final expected = View(
+      classElement,
+      name,
+      fields,
+      query,
+      constructor,
+    );
+    expect(actual, equals(expected));
+  });
+
```
vitusortner, line 68: 💡 We could think about adding some tests that assert the error cases:

```suggestion
test('Throws when processing view without SELECT', () async {
    final classElement = await createClassElement('''
      @DatabaseView("DELETE all from Person")
      class Person {
        final int id;
      
        final String name;
      
        Person(this.id, this.name);
      }
    ''');

    final actual = () => ViewProcessor(classElement, {}).process();

    expect(actual, throwsInvalidGenerationSourceError());
  });

  test(
      'Throws when processing view starting with WITH statement without SELECT',
      () async {
    final classElement = await createClassElement('''
      @DatabaseView("WITH subquery")
      class Person {
        final int id;
      
        final String name;
      
        Person(this.id, this.name);
      }
    ''');

    final actual = () => ViewProcessor(classElement, {}).process();

    expect(actual, throwsInvalidGenerationSourceError());
  });

  test(
      'Throws when processing view starting with WITH statement with only one SELECT',
      () async {
    final classElement = await createClassElement('''
      @DatabaseView("WITH subquery as (SELECT * from otherentity)")
      class Person {
        final int id;
      
        final String name;
      
        Person(this.id, this.name);
      }
    ''');

    final actual = () => ViewProcessor(classElement, {}).process();

    expect(actual, throwsInvalidGenerationSourceError());
  });
```

vitusortner (changes requested):

floor_generator/test/processor/view_processor_test.dart
```diff
@@ -36,6 +36,36 @@ void main() {
...
+      }
+    ''');
+
+    final actual = ViewProcessor(classElement, {}).process();
+
+    const name = 'Person';
+    final fields = classElement.fields
+        .map((fieldElement) => FieldProcessor(fieldElement, null).process())
+        .toList();
+    const query = 'WITH subquery as (SELECT * from otherentity) SELECT subquery.*';
```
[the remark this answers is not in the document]
themadmrj: I don't get why there are errors everywhere now all of a sudden. I'll take a better look tomorrow.

floor_generator/lib/processor/view_processor.dart
```diff
@@ -46,7 +46,24 @@ class ViewProcessor extends QueryableProcessor<View> {
...
+
+  /// whether the string is a common table expression
+  /// followed by a `SELECT` query
+  bool get isCteWithSelect {
+    final lowerCasedString = toLowerCase();
+    return lowerCasedString.trimLeft().startsWith('with') &&
+        'select'.allMatches(lowerCasedString).length >= 2;
+  }
+}
       throw _processorError.missingQuery;
```
vitusortner, line 67: 💡 Oops. It seems like we messed up the file a bit. We can delete this and all following lines.

floor_generator/test/processor/view_processor_test.dart
```diff
@@ -36,6 +36,90 @@ void main() {
     expect(actual, equals(expected));
   });
 
+  test('Process view starting with WITH statement', () async {
```
vitusortner, line 39: 💡 It might be a good idea to run auto format another time on this file.

vitusortner (changes requested):

themadmrj:
The test was cancelled but it seems all tests passed.

themadmrj (commented):
Everything seems A-OK

vitusortner (approved):
👏 thanks for your contribution!

themadmrj:
> 👏 thanks for your contribution!

My pleasure, an Android I'm a huge fan of Room so it's only natural that I think floor's the best database library for Flutter!

Issue #275, opened 2020-03-06
Mapper defaults to the first constructor


jaumzera:
My entities were being fetched with all attributes set to `null` because I had a non-args named constructor (e.g `MyEntity.empty()`) before the all-args one.

I didn't find any mention about this on the docs.

https://github.com/vitusortner/floor/blob/39787be83195c28e46a0ec449d739d9dceea4416/floor_generator/lib/processor/entity_processor.dart#L225

vitusortner:
Good find. We're currently just using the **first** constructor to create in-memory objects. [I'll mention that in the README](https://github.com/vitusortner/floor/pull/302).

Do you have an idea how to improve this behavior?

PR #477, opened 2021-01-07
create object with json serializable support

johnshazhu:
create object add support with json serializable, defalut false (use first constructor)


johnshazhu:
When entity class contains more fields,  it is inconvenient to code its constructor. When create object in memory, is it appropriate 
 to consider use json serializable.

johnshazhu:
This may be link with issue #275

johnshazhu:
@vitusortner If you have some free time, could you review this pull request.

PR #361, opened 2020-06-15
Integrate sqlparser

mqus:
I have now finished an early working version of the integrated sqlparser. There are still some `//todo`s open, most of them related to the typeConverters, because I assume that this PR will not be merged before #318 (though mostly because the query parser is not quite finished).

### Features :
- An even more intelligent generator overall and the foundation for more features
- Query validation for `@DatabaseView` and `@Query`. (#58)
- Type derivation and matching for `@DatabaseView` and `@Query`.
- Dependency and "What entities are affected" derivation, including foreign keys of entities
- Better parameter handling 
  - automatic reordering
  - multiple insertions of the same parameter (#142)
  - better handling of list parameters (fixing sql-injections)
  - removed string-replace functions which could mess up content in string literals
- Outputting simple values in Queries (#200)
- Cheaper streams on DatabaseViews

### TODO:
**Implement some Features/Refactor**
- [x] Mapper for primitive types
- [x] Return type checker
- [x] Restructure  AnalyzeResult creation
- [x] Maybe separate type checker from view converter
- [x] Check view query for variables
- [x] Map bool type to int with typehint in converters
- [x] Make listInsertionPositions a (sorted) list with dedicated type
- [x] Formulate missing errors
- [ ] Check for and add missing errors

**Cleanup**
- [x] Check todo comments for missing functionality and expand this todo list
- [x] Add `@nonNull`/`@nullable` annotations to new functions and check usage
- [x] Add Docstrings to crucial functions

**Integration of other PRs**
- [ ] integrating typeconverters
- [ ] Maybe integrate #359, if it is merged.

**Fixing existing tests**
- [x] integration tests (pretty much already worked without any changes)
- [x] queryadapter unit tests
- [x] generator unit tests

**Adding new tests**
- [ ] integration tests
- [x] queryadapter unit tests
- [ ] generator unit tests
  - [x] Add regression test from #353

**Update README**
- [x] Document subtleties, such as executing a query only up to the first semicolon and not allowing more than 1000 parameter values (both limitations of the sqlite implementations of e.g. android)
- [x] Update Documentation regarding streams on Views

### Small Guide to the PR
I used the library sqlparser from https://github.com/simolus3/moor, like mentioned before.
I added a new subdirectory to `processor` called `query_analyzer`. There are multiple files in there:

[...]

PR #509, opened 2021-03-14
Improved Parameter mapping for query methods

mqus:
As I wrote in #361, the proposed feature set should be done in smaller steps. So I started with this one, which works without integrating an sqlparser for now and seems to come up frequently.

This changeset allows to reference query method parameters in random order and repeatedly. It also adds a small fix to insert Lists via sqlites variable mechanism, too. (and not via string replacement). It should be completely compatible with type converters.

I took the liberty to restructure the query method writer like in #361 to be able to easily insert the changes.

It's probably best to review these changes commit by commit, with the third one being the most complex. Please also advise if I should explain the method/algorithm for parameter matching (processor and writer) better and which details to cover.

I added extensive tests and I could leave all existing integration tests intact, which means that there should be no breakage.

I have multiple next steps in mind and will attempt them once this PR is merged.

### DIFF
--- a/floor_generator/lib/misc/extension/string_extension.dart
+++ b/floor_generator/lib/misc/extension/string_extension.dart
@@ -16,4 +16,40 @@ extension StringExtension on String {
         return this[0].toLowerCase() + substring(1);
     }
   }
+
+  /// Returns a copy of this string having its first letter uppercased, or the
+  /// original string, if it's empty or already starts with a upper case letter.
+  ///
+  /// ```dart
+  /// print('abcd'.capitalize()) // Abcd
+  /// print('Abcd'.capitalize()) // Abcd
+  /// ```
+  String capitalize() {
+    switch (length) {
+      case 0:
+        return this;
+      case 1:
+        return toUpperCase();
+      default:
+        return this[0].toUpperCase() + substring(1);
+    }
+  }
+}
+
+extension NullableStringExtension on String? {
+  /// Converts this string to a literal for
+  /// embedding it into source code strings.
+  ///
+  /// ```dart
+  /// print(null.toLiteral())   // null
+  /// print('Abcd'.toLiteral()) // 'Abcd'
+  /// ```
+  String toLiteral() {
+    if (this == null) {
+      return 'null';
+    } else {
+      //TODO escape correctly
+      return "'$this'";
+    }
+  }
 }
--- a/floor_generator/lib/processor/query_processor.dart
+++ b/floor_generator/lib/processor/query_processor.dart
@@ -17,12 +17,10 @@ class QueryProcessor extends Processor<Query> {
 
   @override
   Query process() {
+    _assertNoNullableParameters();
+
     final listParameters = <ListParameter>[];
-
     final newQuery = _processParameters(listParameters);
-
-    //TODO
-    //_assertNoNamedVarsLeft(newQuery);
 
     return Query(
       newQuery,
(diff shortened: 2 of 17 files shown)


vitusortner:
That's amazing! Thanks for your contribution. I'll look into it as soon as possible.

floor_generator/lib/processor/query_processor.dart
```diff
@@ -0,0 +1,145 @@
...
+      newQuery.write(_query
+          .substring(currentLast, varToken.startPosition)
+          .replaceAll('\n', ' '));
+      final varIndexInMethod = indices[varToken.name];
+      if (varIndexInMethod == null) {
+        throw _processorError.unknownQueryVariable(varToken.name);
+      } else if (varIndexInMethod > 0) {
+        //normal variable/parameter
+        if (varToken.isListVar)
+          throw _processorError.queryArgumentsAndMethodParametersDoNotMatch;
```
vitusortner, line 60: ❓ Will this code path ever be executed? The error message might need to get updated, if so.

floor_generator/lib/processor/query_processor.dart
```diff
@@ -0,0 +1,145 @@
...
+        newQuery.write('?');
+        newQuery.write(varIndexInMethod);
+      } else {
+        //list variable/parameter
+        if (!varToken.isListVar)
+          throw _processorError.queryArgumentsAndMethodParametersDoNotMatch;
```
vitusortner, line 66: See comment above ☝️

floor_generator/lib/processor/query_processor.dart
```diff
@@ -0,0 +1,145 @@
...
+  QueryProcessor(MethodElement methodElement, this._query)
+      : _parameters = methodElement.parameters,
+        _processorError = QueryProcessorError(methodElement);
+
+  @override
+  Query process() {
+    _assertNoNullableParameters();
+
+    final listParameters = <ListParameter>[];
+    final newQuery = _processParameters(listParameters);
```
vitusortner, line 23: 💡 Passing the `listParameters` list into `_processParameters` makes it a bit trickier to follow what's happening to the `listParameters`. Instead, we could make `_processParameters` return a structure that holds both the returned query and the `listParameters`, which is basically `Query`. Thereby, inlining `_processParameters` might help to improve readability.

floor_generator/lib/writer/query_method_writer.dart
```diff
@@ -48,115 +50,148 @@ class QueryMethodWriter implements Writer {
   String _generateMethodBody() {
     final _methodBody = StringBuffer();
 
-    final valueLists = _generateInClauseValueLists();
-    if (valueLists.isNotEmpty) {
-      _methodBody.write(valueLists.join(''));
-    }
+    // generate the variable definitions which will store the sqlite argument
+    // lists, e.g. '?5,?6,?7,?8'. These have to be generated for each call to
+    // the querymethod to accommodate for different list sizes. This is
```
vitusortner, line 55:
```suggestion
    // the query method to accommodate for different list sizes. This is
```

floor_generator/test/processor/query_processor_test.dart
```diff
@@ -0,0 +1,54 @@
...
+    testVarFind(
+        'list_variable3', 'x IN   (:foo)', [VariableToken(':foo', 8, true)]);
+    testVarFind(
+        'list_variable4', 'x In (:fo2o)', [VariableToken(':fo2o', 6, true)]);
+    testVarFind('list_variable5', ':2x iN (:fo2o)',
+        [VariableToken(':2x', 0, false), VariableToken(':fo2o', 8, true)]);
+  });
+}
+
+void testVarFind(
```
vitusortner, line 46: 💡 We could make this function private as its only usages are in this file.

```suggestion
void _testVarFind(
```

vitusortner (dismissed):
This is looking great! Thanks for extensively covering the changes with tests. I just left a few questions and minor suggestions 🚀

floor_generator/lib/misc/extension/string_extension.dart
```diff
@@ -16,4 +16,40 @@ extension StringExtension on String {
...
+  ///
+  /// ```dart
+  /// print(null.toLiteral())   // null
+  /// print('Abcd'.toLiteral()) // 'Abcd'
+  /// ```
+  String toLiteral() {
+    if (this == null) {
+      return 'null';
+    } else {
+      //TODO escape correctly
```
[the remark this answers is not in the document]
mqus: yes, atm this only does the same as before, but in the future this should be able to handle things like converting 
`'\''` to `"'\''"`

floor_generator/lib/processor/query_processor.dart
mqus: this happens if the parameter is a List and the query location looks like it shouldn't accept a list (and the reverse for above). There also is a test for that and the case above.

I agree though, the error message(s) could be improved.

floor_generator/lib/misc/extension/string_extension.dart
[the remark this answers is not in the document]
mqus: (currently it will only do `"'''"` which will generate wrong code)

floor_generator/lib/processor/query_processor.dart
mqus: good point! I'll do that.

floor_generator/lib/misc/extension/string_extension.dart
```diff
@@ -16,4 +16,40 @@ extension StringExtension on String {
...
+
+extension NullableStringExtension on String? {
+  /// Converts this string to a literal for
+  /// embedding it into source code strings.
+  ///
+  /// ```dart
+  /// print(null.toLiteral())   // null
+  /// print('Abcd'.toLiteral()) // 'Abcd'
+  /// ```
+  String toLiteral() {
```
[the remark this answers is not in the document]
mqus: it is currently used for writing out the query parts of the query method body. It should probably also used for the query part of the views. Having looked at that just now, I should change the current "escaping algorithm" a small bit (use `"''' '''"` instead of `"' '"`). This still leaves some holes but they should be rarely used.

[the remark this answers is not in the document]
mqus: scratch that, changing this will blow up this PR rapidly (many tests have to be changed). I think this can be addressed better in a separate PR.

floor_generator/lib/processor/query_processor.dart
mqus: error messages were improved.

floor_generator/lib/processor/query_processor.dart
mqus: error messages were improved.

vitusortner (approved):
Thanks for implementing this long-awaited feature! 🎉

mqus:
Can/Should i merge or is there something left to do?

vitusortner:
> Can/Should i merge or is there something left to do?

Nothing left to do from my perspective. Feel free to merge 🏁

PR #510, opened 2021-03-17
Add support for DateTime

ErliSoares:


dkaera:
DateTime is an Object, so, in my opinion, would better to use a self-created TypeAdapter for the corresponding runtime type.

ErliSoares:
Thanks @dkaera

PR #513, opened 2021-03-17
Add support for uninitialized fields in the constructor

### LINKED ISSUE #512: Add support for uninitialized fields in the constructor
## Entity

```DART
@entity
class MaquinaTipoEntry {
  MaquinaTipoEntry({int? id, this.empresaId, this.syncId, this.nome, this.ativo = true});

  int? id;
  
  int? empresaId;

  int? syncId;

  String? nome;
  
  bool ativo;
  
  int? pessoaInclusaoId;

  int? pessoaEdicaoId;

  DateTime? dataHoraInclusao;

  DateTime? dataHoraUltimaEdicao;
}
```

## DAO

```DART
@dao
abstract class MaquinaTipoDao {
  @Query('SELECT * FROM maquina_tipo WHERE id = :id')
  Future<MaquinaTipoEntry?> findMaquinaTipoById(int id);
}

```

## Generate code

```DART
  @override
  Future<MaquinaTipoEntry?> findMaquinaTipoById(int id) async {
    return _queryAdapter.query('SELECT * FROM maquina_tipo WHERE id = ?',
        arguments: [id],
        mapper: (Map<String, Object?> row) => MaquinaTipoEntry(
            id: row['id'] as int?,
            empresaId: row['empresaId'] as int?,
            syncId: row['syncId'] as int?,
            nome: row['nome'] as String?,
            ativo: (row['ativo'] as int) != 0);
  }
```

## Expected

```DART
  @override
  Future<MaquinaTipoEntry?> findMaquinaTipoById(int id) async {
    return _queryAdapter.query('SELECT * FROM maquina_tipo WHERE id = ?',
        arguments: [id],
        mapper: (Map<String, Object?> row) => MaquinaTipoEntry(
            id: row['id'] as int?,
            empresaId: row['empresaId'] as int?,
            syncId: row['syncId'] as int?,
            nome: row['nome'] as String?,
            ativo: (row['ativo'] as int) != 0)
          ..id = row['id'] as int?
          ..pessoaInclusaoId = row['pessoaInclusaoId'] as int?
          ..pessoaEdicaoId = row['pessoaEdicaoId'] as int?
          ..dataHoraInclusao = row['dataHoraInclusao'] as DateTime?
          ..dataHoraUltimaEdicao = row['dataHoraUltimaEdicao'] as DateTime?);
  }
```

ErliSoares:
Resolve #512


floor_generator/lib/enum_helper_generator.dart
```diff
@@ -0,0 +1,87 @@
+import 'dart:async';
```
vitusortner, line 1: ❓ What functionality does this class provide?

floor_annotation/lib/src/description.dart
```diff
@@ -0,0 +1,7 @@
+/// An annotation used to specify description to show user.
```
vitusortner, line 1: ❓ What's the use-case for this annotation?

_config.yml
```diff
@@ -0,0 +1 @@
+theme: jekyll-theme-slate
```
vitusortner, line 1: 💡 We could consider adding these config files to the repo's .gitignore to not check them into git.

vitusortner (commented):
Thanks for your contribution!
Could you please add some information on which problem we're solving here and how we're solving it? This will help to review these changes.

floor_annotation/lib/src/description.dart
ErliSoares: Sorry for this file, this annotation is from another resource that I am putting to save the enumerates by taking the values of the annotations, without needing a type converter for each enumerate, more details in the other revision

floor_generator/lib/enum_helper_generator.dart
ErliSoares: This code generates an extension for the enuns ones to get the value and description that is provided in the annotations, example below.


### enum

```DART
enum CafeAspecto {
  @Description('Bom')
  @JsonValue(1)
  bom,
  @Description('Regular')    // description to show to user
  @JsonValue(2)                  // value to save database
  regular,
  @Description('Mau')
  @JsonValue(3)
  mau,
}
```

### generated extension

```DART
extension CafeAspectoValueExtension on CafeAspecto {
  int get value {
    switch (this) {
      case CafeAspecto.bom:
        return 1;
      case CafeAspecto.regular:
        return 2;
      case CafeAspecto.mau:
        return 3;
    }
  }

  String get description {
    switch (this) {
      case CafeAspecto.bom:
        return 'Bom';
      case CafeAspecto.regular:
        return 'Regular';
      case CafeAspecto.mau:
        return 'Mau';
    }
  }
}
```

### Example usage

```DART
final value = CafeAspecto.bom.value;              // value is 1
final description = CafeAspecto.bom.description;  //description is Bom
```

I intend to put it to get this value and save it in the database automatically instead of having a type converter for each enumerated

PR #531, opened 2021-04-09
(refactor) Improve escaping by using library

### LINKED ISSUE #352: Can't compare fields against blank

mqus:
This is just a minor improvement for handling issues like #352 properly.

I still have to add an integration test which tests against the specific issue.

Theoretically we could implement this escaping method ourselves but we would have to keep up with dart language changes.

Another option would be to use `'''` and `"""` or even `r'''` but even those can fail in the face of arbitrary queries, so a proper escaping mechanism should be better, since it comes at very little compiletime and no runtime cost.

Closes #352 

- [x] Add an integration test

### DIFF
--- a/example/pubspec.lock
+++ b/example/pubspec.lock
@@ -270,6 +270,13 @@ packages:
       url: "https://pub.dartlang.org"
     source: hosted
     version: "4.0.1"
+  lists:
+    dependency: transitive
+    description:
+      name: lists
+      url: "https://pub.dartlang.org"
+    source: hosted
+    version: "1.0.0"
   logging:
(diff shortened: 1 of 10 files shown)


T-Oner (approved):

vitusortner (approved):
Thanks!

Issue #534, opened 2021-04-12
Cannot set PRAGMA foreign_keys = OFF on migration


cassioseffrin:
As SQLite has some limitations on ALTER TABLE , we usually have to DROP an existing table and create a new one to achieve a simple migration on a table.

Here is the orientations on the oficial docs: http://www.sqlite.org/faq.html:

The basic Steps:
1```
. CREATE TEMPORARY TABLE t1_backup(a,b);
2. INSERT INTO t1_backup SELECT a,b FROM t1;
3. DROP TABLE t1;
4. CREATE TABLE t1(a,b);
5. INSERT INTO t1 SELECT a,b FROM t1_backup;
6. DROP TABLE t1_backup;

``` 


So in my migration script I have tried to do it as described bellow:
```
  await database.execute("PRAGMA foreign_keys = OFF");
  await database.execute("CREATE TABLE IF NOT EXISTS  temp_table AS SELECT * FROM customer");
  await database.execute("DROP TABLE IF EXISTS customer");
  await database.execute("PRAGMA foreign_keys = ON");
```
  
But in fact the first query in my script is not been executed, consequently  throwing an error when the DROP table is called. (787) SQLITE_CONSTRAINT_FOREIGNKEY. It's because my customer table has some references that i need to ignore when I am dropping it

If I manually change the script in auto generated by build_runner **database.g.dart**   setting the **PRAGMA foreign_keys = OFF** the migration works fine, but i cannot set PRAGMA foreign_keys = ON after migration.

```
 Future<sqflite.Database> open(String path, List<Migration> migrations,
      onConfigure: (database) async {
        await database.execute('PRAGMA foreign_keys = OFF');
      },
```
So the issue is: how can I set PRAGMA foreign_keys OFF AND ON in the migration script?

mqus:
Hm. I think this stems from the issue that all migration steps are run within a transaction (to be able to roll them back if something fails) and changing this setting is silently ignored within transactions (see https://www.sqlite.org/foreignkeys.html#fk_enable).

I'm not sure how to change this. We could provide an optional parameter which disables the foreign keys for migrations and switches back afterwards, regardless of migration success (seems to be useful imho).

PS: Maybe unrelated, but I'm interested: what are the limitations you're running into with `ALTER TABLE`?

mqus:
I was slightly mistaken. The reason for the issue is right, but the transaction is opened within sqflite and not in floor, so we can not fix this in floor.

[...]

PR #535, opened 2021-04-14
(feature) Add onConfigure callback

mqus:
Enables a workaround for #534 

I'm not sure how to ~~test~~ write tests for this and would appreciate some Guidance :)

### DIFF
--- a/example/lib/database.g.dart
+++ b/example/lib/database.g.dart
@@ -68,6 +68,7 @@ class _$FlutterDatabase extends FlutterDatabase {
       version: 1,
       onConfigure: (database) async {
         await database.execute('PRAGMA foreign_keys = ON');
+        await callback?.onConfigure?.call(database);
       },
       onOpen: (database) async {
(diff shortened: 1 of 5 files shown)


cassioseffrin:
@mqus Thanks for your efforts on this, this feature will be very useful for migrations!
When this commit is ready for production I will be the first to test it!

cassioseffrin (commented):
I cannot see anything wrong, therefore I think it could be approved. 

I cannot figure out how to cover this change with a test.

dkaera (approved):

vitusortner (approved):
Thanks!

cassioseffrin:
@vitusortner what do you think about having a branch with this feature in old floor 0.19.1 to be compatible with flutter 1x?

I have done it, if you agree with that I can make a PR. I think there are other devs with the same needs. Once flutter 2 (null safety) comes with a lot of refactoring needs and there are many libs that haven't support it yet.

PR #539, opened 2021-04-20
Increase test coverage

mqus:
This PR pushes the test coverage over the magic 90% :)

Some errors from the change method processors were pulled out but the rest is pretty much just more tests.

I tried to tackle the biggest target in the codecov heatmap, `value_objects`, but while the `equals` methods seem simple, the `hashCode` methods are wrong in most cases, because the `hashCode` of dart `Element`s can be different even if the dart elements themselves are equal (`==`). But the value objects also don't really benefit all that much from testing since they only have trivial methods left.

So I took the next best target, error messages. Aside from those there are only two tests (database writer with daoGetters and entity processor with indices) that cover something else. I catched a single issue where the equals method of `Index` was not 100% correct(should only be an internal problem at the very most, if it mattered at all) but nothing apart from that.

There is also one test that I had to set to `skip`  because I could not figure out how to trigger that specific error message (`EntityProcessorError.foreignKeyDoesNotReferenceEntity`). I can also remove that if there is no "solution" ;)

### DIFF
--- a/floor_generator/lib/value_object/index.dart
+++ b/floor_generator/lib/value_object/index.dart
@@ -1,3 +1,5 @@
+import 'package:collection/collection.dart';
+
 class Index {
   final String name;
   final String tableName;
@@ -25,7 +27,7 @@ class Index {
           name == other.name &&
           tableName == other.tableName &&
           unique == other.unique &&
-          columnNames == other.columnNames;
+          const ListEquality<String>().equals(columnNames, other.columnNames);
 
   @override
   int get hashCode =>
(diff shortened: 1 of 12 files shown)


floor_generator/lib/value_object/index.dart
```diff
@@ -25,7 +27,7 @@ class Index {
           name == other.name &&
           tableName == other.tableName &&
           unique == other.unique &&
-          columnNames == other.columnNames;
+          const ListEquality<String>().equals(columnNames, other.columnNames);
```
vitusortner, line 30: 🙏 Well spotted!

vitusortner (approved):
Many thanks for making the effort of writing these tests!

PR #542, opened 2021-04-22
Retain index ordering

### LINKED ISSUE #540: [BUG] - Index columns not in order
The generated index only ever orders the columns by the order the properties appear in the class. NOT the order or the string values of `Index.values`

The workaround would be to reorder the properties in the class to match the order you want them in the index. However, this is not possible when using inheritance.

mqus:
Fixes #540 

This will "decrease" relative coverage because two previously covered lines now disappear, but the actual coverage is unchanged... I think we should change how we specify those coverage thresholds :D

### DIFF
--- a/floor_generator/lib/processor/entity_processor.dart
+++ b/floor_generator/lib/processor/entity_processor.dart
@@ -163,23 +163,20 @@ class EntityProcessor extends QueryableProcessor<Entity> {
           // can't happen as Index.unique is non-nullable
           if (unique == null) throw ArgumentError.notNull();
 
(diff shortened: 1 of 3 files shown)


vitusortner (approved):
Thanks for fixing this!

vitusortner:
> This will "decrease" relative coverage because two previously covered lines now disappear, but the actual coverage is unchanged... I think we should change how we specify those coverage thresholds :D

Codecov has so many unexpected aspects to it. I'll force-merge this PR as everything is covered ✌️

Issue #541, opened 2021-04-22
build_runner 2 not compatible with floor_generator


demos77:
With the new update of build_runner 2.0.0 with null safety floor_generator is not compatible.

`
Because build_runner >=2.0.0 depends on code_builder ^4.0.0 and floor_generator 1.0.1 depends on code_builder ^3.7.0, build_runner >=2.0.0 is incompatible with floor_generator 1.0.1.`

How to do ? Do you prepare a new version ?

vitusortner:
We'll look into updating build_runner as soon as possible. Have you checked if pub's dependency overrides can help? https://dart.dev/tools/pub/dependencies#dependency-overrides

demos77:
With your advice i have trying this, but not work

`dependency_overrides:
  code_builder: ^4.0.0`

proninyaroslav:
@demos77 
You also need to override this:
```
build_config: '>=1.0.0 <1.1.0'
```

demos77:
ok it's work for the moment, the time you upgrade :)

demos77:
When do you make a new version / tag ?

PR #543, opened 2021-04-23
Maintenance/update deps

### LINKED ISSUE #504: Migrate floor_generator's dependencies to null-safe versions
We have recently migrated the codebase to support null-safety (#462). We're now generating null-safe code but floor_generator's internal dependencies aren't exclusively null-safe.

Migrate the dependencies to their null-safe versions once available and make sure to not run tests with `--no-sound-null-safety` anymore (CI).

- [x] build
- [x] build_config
- [x] build_test
- [x] code_builder
- [x] dart_style
- [x] source_gen

Additionally, make sure to get rid of the mockito (5.0.2) dependency override, which was introduced to make floor work with source_gen 1.0.0. The tracking issue for mockito can be found here: https://github.com/dart-lang/mockito/issues/373.

mqus:
Address #541 by updating the dependencies

Afaict this should be the last step of the null-safety migration, so I removed the `--no-sound-null-safety` flag from the CI and it also ran through successfully on my PC. 

Closes #504

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -17,7 +17,7 @@ dependencies:
 
 dev_dependencies:
   floor_generator: ^1.0.1
-  build_runner: ^1.12.2
+  build_runner: ^2.0.0
 ```
 
 ## 2. Create an Entity
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,8 +55,7 @@ jobs:
 
       - name: Run tests
         working-directory: floor_generator
-        # TODO #504 remove --no-sound-null-safety once all dependencies have been migrated
-        run: pub run test_cov --no-sound-null-safety
+        run: pub run test_cov
 
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v1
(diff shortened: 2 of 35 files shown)


mqus:
(not sure which label to use though :D )

demos77 (approved):

docs/getting-started.md
```diff
@@ -17,7 +17,7 @@ dependencies:
 
 dev_dependencies:
   floor_generator: ^1.0.1
-  build_runner: ^1.12.2
+  build_runner: ^2.0.0
```
vitusortner, line 20: 💡 We could also update the `build_runner` versions in `README.md` and `floor/README.md`.

vitusortner (approved):
Thanks for keeping the project's dependencies up-to-date!
I'm currently looking into transferring the package to a verified publisher instead of relying on my personal email address for releases. It'll also simplify having multiple publishers so that all maintainers can create releases.

mqus:
I also noticed that there are some files that still contain the 
```dart
// ignore_for_file: import_of_legacy_library_into_null_safe
```
header, which I can probably remove now

mqus:
> Thanks for keeping the project's dependencies up-to-date!

Only because you made it so easy to do that :D

demos77 (approved):

vitusortner (approved):

PR #552, opened 2021-04-27
Fix query method writer failed when specify the query param in sql

### LINKED ISSUE #551: Query method writer failed when specify the query param in sql
The code:

```dart
@Query("SELECT * FROM task WHERE message = 'test'")
Future<List<Task>> findTaskTestMessage();
```

The log:

```
[SEVERE] floor_generator:floor_generator on lib/database.dart:
An error `FormatterException` occurred while formatting the generated source for
  `package:example/database.dart`
which was output to
  `lib/database.floor.g.part`.
This may indicate an issue in the generator, the input source code, or in the
source formatter.
Could not format because the source could not be parsed:

line 91, column 133 of .: Expected to find ','.
   ╷
91 │ @override Future<List<Task>> findTaskTestMessage() async  { return _queryAdapter.queryList('SELECT * FROM task WHERE message = 'test'', mapper: (Map<String, Object?> row) => Task(row['id'] as int?, row['message'] as String, row['name'] as String)); } 
   │                                                                                                                                     ^^
   ╵
line 91, column 129 of .: Expected to find ','.
   ╷
91 │ @override Future<List<Task>> findTaskTestMessage() async  { return _queryAdapter.queryList('SELECT * FROM task WHERE message = 'test'', mapper: (Map<String, Object?> row) => Task(row['id'] as int?, row['message'] as String, row['name'] as String)); } 
   │                                                                                                                                 ^^^^
   ╵
[INFO] Running build completed, took 433ms

[INFO] Caching finalized dependency graph...
[INFO] Caching finalized dependency graph completed, took 11ms

[SEVERE] Failed after 459ms


```

T-Oner:
Fixes #551

Because single quotes are used when specifying parameters in sql, but the `toLiteral` used in QueryMethodWriter wraps the entire sql with single quotes, resulting in an incorrect string.

### DIFF
--- a/example/lib/database.g.dart
+++ b/example/lib/database.g.dart
@@ -156,6 +156,26 @@ class _$TaskDao extends TaskDao {
   }
 
   @override
+  Stream<List<Task>> findTasksWhereMessageIsEmpty() {
+    return _queryAdapter.queryListStream(
+        "SELECT * FROM task WHERE message IS NULL OR message = ''",
+        mapper: (Map<String, Object?> row) =>
+            Task(row['id'] as int?, row['message'] as String),
+        queryableName: 'Task',
+        isView: false);
+  }
+
+  @override
+  Stream<List<Task>> findTasksWhereMessageIsA() {
+    return _queryAdapter.queryListStream(
+        "SELECT * FROM task WHERE message = 'A'",
+        mapper: (Map<String, Object?> row) =>
+            Task(row['id'] as int?, row['message'] as String),
+        queryableName: 'Task',
+        isView: false);
+  }
+
+  @override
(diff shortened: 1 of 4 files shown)


mqus:
Thank you for your contribution! You're completely right in your analysis but I already proposed the change in #531, which is implementation wise better imho.

Your fix is ok, but not complete, since queries could also contain `"` signs at arbitrary places... my approach simply uses a library which escapes all weird signs, hopefully squashing the problem for all possibilities.

You're invited to review the changes in #531, I'm happy to read your opinions!

T-Oner:
@mqus Glad to hear that you are fixing this problem

Issue #558, opened 2021-05-01
[Floor 1.1.0] Impossible to generate database dart file since latest update


istornz:
Hello, I'm trying to update my project with the latest version of Floor.

My project was not migrated to null safety at the moment, so to continue generating database dart file, I added ```// @dart=2.12``` at the beginning of all my dao files like (workaround found here #507).

Now, when I try to generate my database with ```flutter packages pub run build_runner build --delete-conflicting-outputs```, I get 8 errors like this:

``` shell
[SEVERE] floor_generator:floor_generator on lib/database.dart:
An error `FormatterException` occurred while formatting the generated source for
  `package:my_project/database.dart`
which was output to
  `lib/database.floor.g.part`.
This may indicate an issue in the generator, the input source code, or in the
source formatter.
Could not format because the source could not be parsed:

line 295, column 11 of .: Methods must have an explicit list of parameters.
    ╷
295 │ @override Future<List<GameGenre*>*>* findAllGenres() async  { return _queryAdapter.queryList('SELECT * FROM GameHasGenres', mapper: (Map<String, Object?> row) => GameGenre(identifier: row['identifier'] as int?, name: row['name'] as String?, url: row['url'] as String?, slug: row['slug'] as String?)); }
    │           ^^^^^^
    ╵
line 282, column 18 of .: Expected to find '>'.
    ╷
282 │ @override Future<List<GameGenre*>*>* findAllGenres() async  { return _queryAdapter.queryList('SELECT * FROM GameHasThemes', mapper: (Map<String, Object?> row) => GameGenre(identifier: row['identifier'] as int?, name: row['name'] as String?, url: row['url'] as String?, slug: row['slug'] as String?)); }
```
(log not complete)

Is it related to #531 & #551 ?

Any help would be greatly appreciated !
Thanks in advance 👍

mqus:
`Future<List<GameGenre*>*>*` why the `*`s? Maybe I don't know this feature but dart has no pointer type afaik. What did you try to express?

istornz:
@mqus thanks for fast reply !
I don't have any pointer in my code... This is very strange, it seems floor add it when generating :/

Here is my dao:

```
@dao
abstract class GenreDao {
  @Query('SELECT * FROM Genre')
  Future<List<GameGenre>> findAllGenres();

  // [...]
}
```

mqus:
[...]

PR #559, opened 2021-05-01
Bugfix/nullable transaction return

### LINKED ISSUE #557: Transaction: a nullable return type becomes not nullable in the method's implementation
Step to reproduce the issue:
- add this simple method to the TaskDao class in the official [example](https://github.com/vitusortner/floor/blob/develop/example/lib/task_dao.dart)

  ```dart
  @transaction
   Future<Task?> test(int id) {
     return findTaskById(id);
   }
  ```
- run `flutter pub run build_runner build`

Generated code:
```dart
  @override
  Future<Task> test(int id) async {
    if (database is sqflite.Transaction) {
      return super.test(id);
    } else {
      return (database as sqflite.Database)
          .transaction<Task>((transaction) async {
        final transactionDatabase = _$FlutterDatabase(changeListener)
          ..database = transaction;
        return transactionDatabase.taskDao.test(id);
      });
    }
  }
```

This code doesn't compile because the method is declared with a not nullable return type, and also `sqflite.Database.transaction<Task>(..)` (in the else branch) is invoked with a not nullable parametric type

mqus:
Fixes #557 and maybe adresses #558

### DIFF
--- a/floor_generator/lib/writer/transaction_method_writer.dart
+++ b/floor_generator/lib/writer/transaction_method_writer.dart
@@ -14,7 +14,7 @@ class TransactionMethodWriter implements Writer {
     return Method((builder) => builder
       ..annotations.add(overrideAnnotationExpression)
       ..returns = refer(method.returnType.getDisplayString(
-        withNullability: false,
+        withNullability: true,
       ))
       ..name = method.name
       ..requiredParameters.addAll(_generateParameters())
@@ -47,7 +47,7 @@ class TransactionMethodWriter implements Writer {
       return Parameter((builder) => builder
         ..name = parameter.name
         ..type = refer(parameter.type.getDisplayString(
-          withNullability: false,
+          withNullability: true,
         )));
     }).toList();
   }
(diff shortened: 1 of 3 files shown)


floor_generator/lib/writer/transaction_method_writer.dart
```diff
@@ -14,7 +14,7 @@ class TransactionMethodWriter implements Writer {
     return Method((builder) => builder
       ..annotations.add(overrideAnnotationExpression)
       ..returns = refer(method.returnType.getDisplayString(
-        withNullability: false,
+        withNullability: true,
```
vitusortner, line 17: ❓ Do we also get support for the case described in https://github.com/vitusortner/floor/issues/557 where we don't return a nullable `Future` but a `Future` wrapping a nullable type (e.g `Future<Task?>`)? If so, could you add another test to verify the behavior?

vitusortner (commented):
Thanks for the quick fix!

floor_generator/lib/writer/transaction_method_writer.dart
mqus: I think we had that support already, but I'll add a test.

vitusortner (approved):

xThalia:
I've tried to return Future<Object?> from transaction but it still causes an error in generated db file. Will there be support for return Future with nullable object?

PR #634, opened 2022-01-18
Support latest analyzer

paricleu:
Updates analyzer to ^3.0.0 to ensure compatibility with other dependencies.

### DIFF
--- a/example/android/app/build.gradle
+++ b/example/android/app/build.gradle
@@ -25,20 +25,20 @@ apply plugin: 'com.android.application'
 apply from: "$flutterRoot/packages/flutter_tools/gradle/flutter.gradle"
 
 android {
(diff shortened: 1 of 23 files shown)


paricleu:
I tried fiddling with it and managed to get a workable state now. 
The tests run fine and I tested using the analyzer plugin in the example app with version 3.+. While doing that I also updated the example app to use the v2 embedding and newer gradle versions.

vitusortner:
awesome, thanks! feel free to update the example to use analyzer 3 in this pr as well. also, could you fix the linted issues [here](https://github.com/vitusortner/floor/runs/4860091510?check_suite_focus=true)? once this is done, let's merge the pr 🙌

vitusortner:
this is looking great, thanks!

PR #685, opened 2022-08-04
Fix null pointer exception during the generation queries which return…

dkaera:
Recreated PR because the previous one was mistakenly updated during the solving conflicts.

### DIFF
--- a/example/pubspec.lock
+++ b/example/pubspec.lock
@@ -7,21 +7,21 @@ packages:
       name: _fe_analyzer_shared
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "40.0.0"
+    version: "43.0.0"
   analyzer:
     dependency: "direct dev"
     description:
       name: analyzer
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "4.1.0"
+    version: "4.3.1"
   args:
     dependency: transitive
     description:
       name: args
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "2.3.0"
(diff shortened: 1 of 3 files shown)


example/pubspec.lock
```diff
@@ -505,14 +505,14 @@ packages:
       name: web_socket_channel
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "2.1.0"
+    version: "2.2.0"
```
mqus, line 508: Nit: Maybe separate the updated pubspec into another commit next time, since it isn't really necessary for this change. But this is really a minor issue.

mqus (approved):

example/pubspec.lock
dkaera: agree

To be honest, I think `pubspec.lock` should be added to `gitignore` since it is a generated pub file. 

Yes, the dart documentation also [mentions this](https://dart.dev/guides/libraries/private-files#the-rules). So we can remove it from the example project but must leave it in packages.

mqus:
```
pubspec.lock  # Except for application packages
```
I interpret this to mean that it should be .gitignore'd for libraries but not for applications. Technically, the example is an application, not a library.

dkaera: ah, yeah, you're right 😂

ThomasMiddel (approved):

ThomasMiddel (approved):

memoriasIT (approved):

PR #730, opened 2023-01-27
feat: Web support

mirland:
Right now [sqflite_common_ffi_web](https://pub.dev/packages/sqflite_common_ffi_web) is in an experimental state, but if we add the factory, we can support the web optionally.

What do you think?

Currently, floor doesn't support web and we cannot update the factory creator, so if we want to support web we have to change the generated code (or duplicate code), and it's not a good idea.
This pr adds the web factory


dkaera:
wow, looks good, did you test it out?

mirland:
Yes doing these changes manually I could make my application work. I tested in [this branch](https://github.com/xmartlabs/fluttips/tree/feature/web-support), but I did floor changes just locally

dkaera:
> Yes doing these changes manually I could make my application work. I tested in [this branch](https://github.com/xmartlabs/fluttips/tree/feature/web-support), but I did floor changes just locally

awesome, I'll prepare the project update PR, because the current one contains a lot of changes in the `.lock` file, which doesn't related to the main change.

dkaera (dismissed):

dkaera:
Hey @mirland
`ubuntu` job broke, I fixed it in my PR
let's merge your changes after mine

ThomasMiddel (approved):

mirland:
Hi @dkaera, do you have any feedback for my PR? I don't know if I can do something to merge it

mirland:
> the coverage check blocks this PR, so you need to add the corresponding test for the newly added condition first

Hi @dkaera , can you suggest to me how can I do that, in this case, there's no test to do, I mean, if we want to do that we should like mock the platform and check the factory. I checked your tests and that method seems to be not tested and there's not something similar

mirland:
> the coverage check blocks this PR, so you need to add the corresponding test for the newly added condition first

Hi @dkaera , can you suggest to me how can I do that, in this case, there's no test to do, I mean, if we want to do that we should like mock the platform and check the factory. I checked your tests and that method seems to be not tested and there's not something similar

mirland:
@dkaera can you suggest to me what kind of test should I include?

dkaera:
> @dkaera can you suggest to me what kind of test should I include?

ah, I'm sorry, I can't help you now, because of the heavy workload.

mirland:
Ok, np, no rush. Then when you have time, please check it because I changed only 3 lines and I think I cannot do a test for that.

image

Thanks!

dkaera:
> Ok, np, no rush. Then when you have time, please check it because I changed only 3 lines and I think I cannot do a test for that.
> 
> image
> 
> Thanks!

I think we can mock `Platform` value and verify what the instance was factory created.

stefanosiano:
any news on this? 
No rush, i'd just like to try floor instead of sqflite in my project, which runs on desktop and web

hendrikvdkaaden:
I really appreciate this feature and would like to continue working with it. However, I encountered an issue when launching the example project on the web using your branch. The error message displayed is:

-----------------------------------------------------------------------------------------
An error occurred while initializing the web worker.
This is likely due to a failure to find the worker javascript file at sqflite_sw.js

Please check the documentation at https://github.com/tekartik/sqflite/tree/master/packages_web/sqflite_common_ffi_web#setup-binaries to setup the needed binaries.

Error: SqfliteFfiWebException()
-----------------------------------------------------------------------------------------

Do you have any insights into the cause of this problem?

stephanmantel:
> > Ok, np, no rush. Then when you have time, please check it because I changed only 3 lines and I think I cannot do a test for that.
> > image
> > Thanks!
> 
> I think we can mock `Platform` value and verify what the instance was factory created.

Is it possible to mock Platform and kIsWeb? I found no support for static mocking using Mockito.

We could also resort to a wrapper object, but I feel like that misses the point a bit, as the real logic you'd want to test is mocked away. What do you guys think, is is necessary to test this?

dkaera:
Hey guys
With pr #815 issues should sorted
Lets push it to development brunch then we can merge @mirland changes to support web 
I tested it locally - everything works fine 💪💪💪

hendrikvdkaaden:
Is this PR still relevant, since it's merged to #819?

dkaera:
@hendrikvdkaaden yes we can close this PR

dkaera:
the changes were picked up and integrated separately

PR #764, opened 2023-07-03
Add TaskStatus.open as status when creating a new task

### LINKED ISSUE #763: In the example, a task is created with status: null instead of open
This means the filter in the example is set to 'open', new tasks will not be shown.

When the user swipes to update the status, the status will jump to `inProgress`, thus skipping `open`.

SEGVeenstra:
Fixes #763


stephanmantel:
Even though I like most of the formatting changes, it got me thinking is there an agreed upon style guide for this project? And if not, should there be?

stephanmantel (dismissed):

stephanmantel (approved):

PR #781, opened 2023-09-19
fixed issue in query adapter's queryList where CompoundSelectStatemen…

mthongvanh:
…t is not executed

queryList does not allow compound select statements to be executed. UNION queries throw state errors. 
updating the rootNode's type check to BaseSelectStatement instead of SelectStatement fixes the issue.

### DIFF
--- a/floor/lib/src/adapter/query_adapter.dart
+++ b/floor/lib/src/adapter/query_adapter.dart
@@ -44,7 +44,7 @@ class QueryAdapter {
   }) async {
     final rootNode = _parseRootNode(sql);
 
-    if (rootNode is SelectStatement) {
+    if (rootNode is BaseSelectStatement) {
       return _database
           .rawQuery(sql, arguments)
           .then((rows) => rows.map((row) => mapper(row)).toList());


mthongvanh:
Hi Dmytro,
There's not currently an open issue. We just saw an issue recently after
updating our fork.

dkaera:
> Hi Dmytro, There's not currently an open issue. We just saw an issue recently after updating our fork. Should I create an issue first?
Please add a sample code that does not work in the current Floor version and a relevant explanation of the issue. 
It'll help reviewers understand what problem this PR solves.

mthongvanh:
_Floor version 1.4.2
Dart 2.19.6
Flutter 3.7.12_

**Issue:**
When performing a `queryList` with a raw SELECT UNION directly with `QueryAdapter`, the query fails and throws a StateError. Since `sqlparser` returns the rootNode as a `CompoundSelectStatement`, the type check for `SelectStatement` fails. 

**Fix:**
To fix the issue and allow other types of SELECT statements, it should instead check for the `BaseSelectStatement` type which is extended by both `SelectStatement` and `CompoundSelectStatement`.

**Sample Code**
The code below throws a StateError when added to the floor example project.

```
Future<void> main() async {
  WidgetsFlutterBinding.ensureInitialized();

  final database = await $FloorFlutterDatabase
      .databaseBuilder('flutter_database.db')
      .build();
  final dao = database.taskDao;

  var qa = QueryAdapter(database.database);
  var list = await qa.queryList(
      'SELECT message FROM `Task` UNION SELECT message FROM `Task`',
      mapper: (e) => Task.optional(message: e['message'] as String?));
  
  runApp(FloorApp(dao));
}
```

dkaera:
@mthongvanh 
thank you for details
yes, I can see the problem, but we need to verify if we're breaking anything with this change.
Meanwhile, Floor now supports Dart types, so the following approach can be used to retrieve the list of messages in the DAO.
```dart
@Query('SELECT message FROM Task UNION SELECT message FROM Task')
Future<List<String>> queryMessages();
```

PR #807, opened 2024-02-27
Updated dependencies

hendrikvdkaaden:
Updated the dependencies version to the latest version.


stephanmantel (approved):

stephanmantel:
Could you commit an empty commit to re-trigger the checks?

floor/lib/src/sqflite_database_factory.dart
```diff
@@ -1,6 +1,7 @@
 import 'dart:io';
 
 import 'package:path/path.dart';
+// ignore: unnecessary_import
 import 'package:sqflite/sqflite.dart';
```
SEGVeenstra, line 5: If it's unnecessary, can it not just be removed?

SEGVeenstra (approved):
I'm not very experienced with the code base, but for what I can tell it looks good to me with just one question in the comments.
If you don't have any urgency you might want to get a second opinion from @stephanmantel next week.

floor/lib/src/sqflite_database_factory.dart
hendrikvdkaaden: The import is necessary. I added the ignore because when we run the build_runner, it indicates that it's an unnecessary import, causing the pipeline to fail.

PR #808, opened 2024-02-29
All test in floor_generator run now

hendrikvdkaaden:
The command flutter test did not work. To fix this, I updated all the dependencies. After doing this, some problems arose in the fakes.dart and string_extension.dart files, which I also fixed. Additionally, I corrected some tests that were not accurate. To run the tests, you need to use flutter pub run test.

### DIFF
--- a/floor_annotation/lib/src/type_converters.dart
+++ b/floor_annotation/lib/src/type_converters.dart
@@ -24,5 +24,5 @@ class TypeConverters {
   /// The list of type converter classes.
   final List<Type> value;
 
-  const TypeConverters(this.value);
+  const TypeConverters({this.value = const []});
 }
--- a/floor_annotation/lib/src/type_converters.dart
+++ b/floor_annotation/lib/src/type_converters.dart
@@ -22,7 +22,7 @@ import 'package:meta/meta.dart';
 @experimental
 class TypeConverters {
   /// The list of type converter classes.
-  final List<Type> value;
+  final List<Type>? value;
 
-  const TypeConverters({this.value = const []});
+  const TypeConverters(this.value);
 }
--- a/floor_annotation/lib/src/type_converters.dart
+++ b/floor_annotation/lib/src/type_converters.dart
@@ -22,7 +22,7 @@ import 'package:meta/meta.dart';
 @experimental
 class TypeConverters {
   /// The list of type converter classes.
-  final List<Type>? value;
+  final List<Type> value;
 
   const TypeConverters(this.value);
 }
--- a/floor_generator/lib/misc/extension/string_extension.dart
+++ b/floor_generator/lib/misc/extension/string_extension.dart
@@ -55,7 +55,7 @@ extension NullableStringExtension on String? {
     if (this == null) {
       return 'null';
     } else {
-      return "'${escape(this!)}'";
+      return "'${this!.toEscaped()}'";
     }
   }
 }
--- a/floor_generator/test/fakes.dart
+++ b/floor_generator/test/fakes.dart
@@ -258,7 +258,6 @@ class FakeClassElement implements ClassElement {
   @override
   Source get source => throw UnimplementedError();
 
-  @override
   List<InterfaceType> get superclassConstraints => throw UnimplementedError();
 
   @override
@@ -302,7 +301,6 @@ class FakeClassElement implements ClassElement {
   // TODO: implement augmented
   AugmentedClassElement get augmented => throw UnimplementedError();
 
-  @override
   // TODO: implement enclosingElement3
   CompilationUnitElement get enclosingElement3 => throw UnimplementedError();
 
@@ -310,7 +308,6 @@ class FakeClassElement implements ClassElement {
   // TODO: implement hasMustBeOverridden
   bool get hasMustBeOverridden => throw UnimplementedError();
 
-  @override
   bool isAccessibleIn2(LibraryElement library) {
     // TODO: implement isAccessibleIn2
     throw UnimplementedError();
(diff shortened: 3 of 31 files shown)


floor_annotation/lib/src/type_converters.dart
```diff
@@ -22,7 +22,7 @@ import 'package:meta/meta.dart';
 @experimental
 class TypeConverters {
   /// The list of type converter classes.
-  final List<Type> value;
+  final List<Type>? value;
```
stephanmantel, line 25: Why is this nullable? I have the feeling it makes not much sense to use this annotation if you do not have any type converters. In other words, the list will always have at least one element and will never need to be null

floor_generator/lib/misc/extension/string_extension.dart
```diff
@@ -55,7 +55,7 @@ extension NullableStringExtension on String? {
     if (this == null) {
       return 'null';
     } else {
-      return "'${escape(this!)}'";
+      return "'${this!.toEscaped()}'";
```
stephanmantel, line 58: I think the entire function body can be simpler: we can use this?.toEscaped() and use null-coalessing to return the literal 'null'

stephanmantel (changes requested):

floor_generator/test/fakes.dart
```diff
@@ -258,7 +258,6 @@ class FakeClassElement implements ClassElement {
   @override
   Source get source => throw UnimplementedError();
 
-  @override
   List<InterfaceType> get superclassConstraints => throw UnimplementedError();
```
dkaera, line 261: let's remove it if ClassElement doesn't contain it anymore

floor_generator/test/fakes.dart
```diff
@@ -298,23 +297,17 @@ class FakeClassElement implements ClassElement {
...
-  // TODO: implement augmentation
-  ClassAugmentationElement? get augmentation => throw UnimplementedError();
-
   @override
   // TODO: implement augmented
   AugmentedClassElement get augmented => throw UnimplementedError();
 
-  @override
   // TODO: implement enclosingElement3
   CompilationUnitElement get enclosingElement3 => throw UnimplementedError();
```
dkaera, line 305: remove method

floor_generator/test/fakes.dart
```diff
@@ -298,23 +297,17 @@ class FakeClassElement implements ClassElement {
...
 
   @override
   // TODO: implement hasMustBeOverridden
   bool get hasMustBeOverridden => throw UnimplementedError();
 
-  @override
   bool isAccessibleIn2(LibraryElement library) {
     // TODO: implement isAccessibleIn2
     throw UnimplementedError();
```
dkaera, line 313: remove method

floor_generator/test/fakes.dart
```diff
@@ -581,19 +602,13 @@ class FakeFieldElement implements FieldElement {
...
   Element get enclosingElement3 => throw UnimplementedError();
 
   @override
   // TODO: implement hasMustBeOverridden
   bool get hasMustBeOverridden => throw UnimplementedError();
 
-  @override
   bool isAccessibleIn2(LibraryElement library) {
     // TODO: implement isAccessibleIn2
     throw UnimplementedError();
```
dkaera, line 614: remove method

dkaera (changes requested):
amazing work
left a few comments to find out what benefits these changes bring

floor_generator/test/misc/extension/type_converter_element_extension_test.dart
```diff
@@ -37,19 +37,6 @@ void main() {
...
-      final element = await '''
-        @TypeConverters(null)
-        abstract class Foo {}
-      '''
-          .asClassElement();
-
-      final actual = () => element.getTypeConverters(typeConverterScope);
-
-      expect(actual, throwsProcessorError());
-    });
```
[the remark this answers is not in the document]
stephanmantel: Hi, thanks for your input :D. 
The author and I spoke briefly about this offline and this is what I thought:
The TypeConverters annotation cannot take a null value anymore. (Which in my opinion makes sense, because if you have no type converters, you can just not use the annotation, so I expect there always to be at least one element in the list).
Therefore, with a null value, it won't compile so there is no need for a test if it already fails at compile time.
But of course, please correct me if this view is too limited.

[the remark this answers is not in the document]
dkaera: ok, got it
But in any case, if we're talking about the code generation phase rather than compile time, we need to be sure that this situation causes the correct error.
This was the main idea of this test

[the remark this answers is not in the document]
stephanmantel: Thanks for your input, I understand, lets get those tests back in and get them to expect the right error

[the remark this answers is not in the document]
hendrikvdkaaden: I put the test back and it expect the right error now.

floor_generator/test/misc/extension/type_converter_element_extension_test.dart
```diff
@@ -37,7 +37,7 @@ void main() {
       expect(actual, equals({expected}));
     });
 
-    test('throws error when null in annotation', () async {
+    test('throws error when empty list in annotation', () async {
```
stephanmantel, line 40: This is incorrect, this test checks null, not empty list

floor_generator/test/test_utils.dart
```diff
@@ -157,6 +157,12 @@ Matcher throwsProcessorError([
   }
 }
 
+Matcher throwsUnresolvedAnnotationException() {
+  return throwsA(isA<UnresolvedAnnotationException>());
+}
+
```
stephanmantel, line 163: As indicated by the failing CI, you have too many empty lines here

stephanmantel (approved):

PR #809, opened 2024-03-07
Update ci.yml

0biWanKenobi:
Should make #807 work, the pipeline reports failures related to using a Dart version that's incompatible


stephanmantel (approved):

dkaera:
I would not recommend merging PR without successfully passing all the checks.

Issue #811, opened 2024-03-08
Imports collision


alex-melnyk:
Using Flutter SDK **3.7.12** and **3.10.6** I faced an imports collision:

```console
00:13 +0: loading /builds/mycompany/myproj/test/database/entity/clock_in_entity_test.dart

.pub-cache/hosted/pub.dev/floor-1.4.2/lib/src/sqflite_database_factory.dart:10:12:
Error: 'databaseFactory' is imported from both 'package:sqflite/src/factory_impl.dart' and
'package:sqflite_common/src/sqflite_database_factory.dart'.
    return databaseFactory;
           ^^^^^^^^^^^^^^^
```

stephanmantel:
This should be fixed with version 1.5.0, please reopen if it isn't

PR #812, opened 2024-03-08
#811 fix imports collision

alex-melnyk:
#811 Added import names to avoid import collision.

### DIFF
--- a/floor_generator/pubspec.yaml
+++ b/floor_generator/pubspec.yaml
@@ -17,7 +17,8 @@ dependencies:
   build_config: ^1.0.0
   code_builder: ^4.1.0
   collection: ^1.15.0
-  floor_annotation: 1.4.2
+  floor_annotation:
+    path: ../floor_annotation
   meta: ^1.7.0
   source_gen: ^1.2.2
   strings: ^0.2.1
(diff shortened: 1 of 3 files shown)


floor_generator/pubspec.yaml
```diff
@@ -17,8 +17,7 @@ dependencies:
   build_config: ^1.0.0
   code_builder: ^4.1.0
   collection: ^1.15.0
-  floor_annotation:
-    path: ../floor_annotation
```
stephanmantel, line 21: For development purposes we should point these to the local one, so these should not be changed

stephanmantel:
Good change, however we are currently working on web support (#526) which also adds named imports, so I wonder if this one is still necessary after that. So I suggest to get that one in first

floor_generator/pubspec.yaml
alex-melnyk: Yeah, I see it is comfortable to use relative imports in the development stage, but looking forward, in terms of using the plugin with my version until the next version comes out, it is better to change that coz there is not way to import  plugin from the GIT directly because the relative imports.

alex-melnyk: > For development purposes we should point these to the local one, so these should not be changed

Yeah, I see it is comfortable to use relative imports in the development stage, but looking forward, in terms of using the plugin with my version until the next version comes out, it is better to change that coz there is no way to import plugin from the GIT directly because the relative imports.

stephanmantel:
No longer relevant, closed

PR #821, opened 2024-04-02
Throw error when encountering `InvalidStatement`

### LINKED ISSUE #820: Invalid query (failing to parse) does not throw an error
In our app a wrong SQL statement in the form of `DELETE * FROM foo` (where there should be no `*` in the statement) snuck into a `@Query` annotation.

The underlying `sqlparser` throws a `ParsingError` in

https://github.com/simolus3/drift/blob/sqlparser-0.27.0/sqlparser/lib/src/reader/parser.dart#L141

which then gets converted into a `InvalidStatement` return value on a higher level (instead of throwing the error higher up):

https://github.com/simolus3/drift/blob/sqlparser-0.27.0/sqlparser/lib/src/reader/parser.dart#L185

But then in `floor`'s query execution this type of "statement" is silently ignored and the error is never reported:

https://github.com/pinchbv/floor/blob/1.4.2/floor/lib/src/adapter/query_adapter.dart#L141-L152

I would suggest that the type-switch would be extended to handle `InvalidStatement` explicitly (forwarding the error) and then have a general fallback to throw on unknown statement types (as we can't exhaustively switch over them, as the base class is not `sealed`).

Does that sound a reasonable extension?

tp:
Fixes #820


stephanmantel (dismissed):

stephanmantel:
Yes, looks good to me, can you solve the conflicts?

tp:
@stephanmantel Rebased on the `develop` branch here.

tp:
@stephanmantel Anything I can further help with to move this one forward?

stephanmantel (approved):

PR #822, opened 2024-04-03
feat: add external path support

lucasdidur:
Add support for Floor to set a external path to load/save database. App need to have storage permission managed by Permissions Manager.

```dart
  Future<AppDatabase> get appDatabase async {
    String externalPath = await ExternalPath.getExternalStoragePublicDirectory(
      ExternalPath.DIRECTORY_DOCUMENTS,
    );
    var dir = Directory(p.join(externalPath, 'My App'));
    if (!dir.existsSync()) {
      dir.create(recursive: true);
    }

    return $FloorAppDatabase.databaseBuilder(p.join(dir.path, DATABASE_NAME)).build();
  }
```

### DIFF
--- a/floor/lib/src/sqflite_database_factory.dart
+++ b/floor/lib/src/sqflite_database_factory.dart
@@ -19,7 +19,15 @@ final DatabaseFactory sqfliteDatabaseFactory = () {
 
 extension DatabaseFactoryExtension on DatabaseFactory {
   Future<String> getDatabasePath(final String name) async {
-    final databasesPath = await this.getDatabasesPath();
-    return join(databasesPath, name);
+    late String databasesPath;
+
+    final file = File(name);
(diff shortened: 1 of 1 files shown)


ryanheise:
I like this PR. For most apps, the database should probably not be stored where this package wants to store it - `path_provider.getApplicationSupportDirectory()` would be better in most cases. The sqflite documentation says as much, and suggests not using `getDatabasesPath()` in favour of using `path_provider` directly, and constructing your own database path to pass into the API.

dfdgsdfg:
@stephanmantel 
Seems good to enough to merge. What do you think?

ryanheise:
Actually, you can just do this:

```
sqfliteDatabaseFactory.setDatabasesPath(...);
```

PR #829, opened 2024-05-21
Support embedded objects

hendrikvdkaaden:
This pull request introduces support for embedded objects in the Floor library. Initially, I referred to pull request #565 but due to the extensive outdated changes, I decided to start fresh to avoid complex merge conflicts.


example/lib/timestamp.dart
```diff
@@ -0,0 +1,32 @@
+import 'package:floor/floor.dart';
+import 'package:intl/intl.dart';
+
+@Embed()
```
stephanmantel, line 4: To remain constistent with the @entity annotation, can you add a const for this?

```
/// Marks a class as a database entity (table).
const entity = Entity();
```

This way, only one embed object will be created in memory and be reused, which improved memory usage. Of course, it is only minimal,but still a best practice

example/lib/type_converter.dart
```diff
@@ -24,3 +24,15 @@ class TaskTypeConverter extends TypeConverter<TaskType?, String?> {
     return value?.name;
   }
 }
+
+class TaskStatusConverter extends TypeConverter<TaskStatus?, String?> {
```
stephanmantel: This one is not needed, because TaskStatus is an enum

example/test/main_test.dart
```diff
@@ -14,6 +14,7 @@ void main() {
   });
 
   tearDown(() async {
+    await Future.delayed(const Duration(milliseconds: 100));
```
stephanmantel, line 18: As we found out, this is necessary to pass the tests, otherwise the database is closed before the test is finished. I do not really like this solution, but it is okay for now. But can you add a comment to describe its necesity?

example/lib/timestamp.dart
stephanmantel: Oh, my bad, I see it already exists even, so please use it

example/lib/timestamp.dart
```diff
@@ -0,0 +1,32 @@
...
+      other is Timestamp &&
+          runtimeType == other.runtimeType &&
+          createdAt == other.createdAt &&
+          updatedAt == other.updatedAt;
+
+  @override
+  int get hashCode => createdAt.hashCode ^ updatedAt.hashCode;
+
+  @override
+  String toString() {
```
stephanmantel: I don't really like using the toString function to for user-facing strings. I think it would be better to keep toString for technical things and introduce a new function, something like: format() to get a pretty user-facing String

stephanmantel:
This is already a good thing, but I noticed that it does not work if the embeded attribute is nullable. Can you add support for that?

kylechandev (approved):

PR #855, opened 2025-03-07
Updates to support Dart 3.7.0 (Flutter 3.29.X)

jwyrembelski:
We've begun updating our Flutter project to use 3.29.X and find that `floor` isn't compatible. Fundamentally, this is a duplicate concern as raised in [New version of source_gen](https://github.com/pinchbv/floor/issues/852).

I started by retreating to a version of Dart that would yield passing unit tests and found that Dart 3.2.6 (as shipped with Flutter 3.19.6) gave me what I wanted. 

With this baseline in hand, I upgraded all dependencies to the latest (or nearly the latest) to support Dart 3.7.0 (as shipped with Flutter 3.29.0). Along the way, I had to drop test_cov (RIP). 

After achieving passing unit tests with the updated dependencies, I attempted to run the example app and found it so outdated that I needed to replace the build files. I added macOS and Windows but removed the web implementation (temporarily?).


toshiossada:
need this update

toshiossada (approved):
LGFM

boginni:
This update is essential

codingwithsaeed:
Please consider merging this pull request. I have a ton of projects using Floor that I've held back from upgrading because of it.

for5million:
need this，thanks everyone!

Vincent-dejong:
Any update on this? Also holding me back on a project... 😢

PR #862, opened 2025-08-19
Dependencies migrated to new versions

BetterB0y:
From PR: https://github.com/pinchbv/floor/pull/855
I've made necessary changes to use new version of build to be still compatible with other libraries.


Vincent-dejong (approved):
LGTM

